PR #59, opened 2020-11-25
feat(gorilla/mux): adds support for gorilla/mux.

jcchavezs:
This PR adds support for `gorilla/mux`.

TODO: verify body/headers and being recorded as expected.

### DIFF
--- a/instrumentation/opentelemetry/github.com/gorilla/hypermux/mux.go
+++ b/instrumentation/opentelemetry/github.com/gorilla/hypermux/mux.go
@@ -12,14 +12,18 @@ import (
 
 func spanNameFormatter(operation string, r *http.Request) (spanName string) {
 	route := mux.CurrentRoute(r)
-	if route == nil {
-		return
+	if route != nil {
+		var err error
+		spanName, err = route.GetPathTemplate()
+		if err != nil {
+			spanName, _ = route.GetPathRegexp()
+		}
 	}
 
-	var err error
-	spanName, err = route.GetPathTemplate()
-	if err != nil {
-		spanName, _ = route.GetPathRegexp()
+	if spanName == "" {
+		// if somehow retrieving the path template or path regexp fails, we still
+		// want to use the method as fallback.
+		spanName = r.Method
 	}
 
 	return
--- a/instrumentation/opentelemetry/github.com/gorilla/hypermux/mux.go
+++ b/instrumentation/opentelemetry/github.com/gorilla/hypermux/mux.go
@@ -1,0 +1,38 @@
+package hypermux
+
+import (
+	"net/http"
+
+	"github.com/gorilla/mux"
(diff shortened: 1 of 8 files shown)


instrumentation/opentelemetry/github.com/gorilla/hypermux/mux.go
```diff
@@ -0,0 +1,38 @@
...
+	spanName, err = route.GetPathTemplate()
+	if err != nil {
+		spanName, err = route.GetPathRegexp()
+	}
+
+	return
+}
+
+// NewMiddleware sets up a handler to start tracing the incoming requests.
+func NewMiddleware() mux.MiddlewareFunc {
```
jcchavezs, line 29: This is the actual Gorilla mux instrumentation.

instrumentation/opentelemetry/github.com/gorilla/hypermux/mux_test.go
```diff
@@ -0,0 +1,62 @@
...
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		t.Errorf("unexpected error: %v", err)
+	}
+
+	if want, have := 202, res.StatusCode; want != have {
+		t.Errorf("unexpected status code, want %q, have %q", want, have)
+	}
+
+	spans := flusher()
```
jcchavezs, line 58: We also need to verify body/headers are being recorded.

instrumentation/opentelemetry/github.com/gorilla/hypermux/mux.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func spanNameFormatter(operation string, r *http.Request) (spanName string) {
+	route := mux.CurrentRoute(r)
+	if route == nil {
+		return
+	}
+
+	var err error
+	spanName, err = route.GetPathTemplate()
```
tim-mwangi, line 20: spanName is always nil when err is not? In that case it will be modified, no?

jcchavezs: if `err != nil`, spanName is `""`, still we need it as types must be respected.

tim-mwangi: In that case to we want an empty spanName? Shouldn't we return the value that was passed in to this method?

jcchavezs: Good point. I will default to `request.Method`.

tim-mwangi (approved):

instrumentation/opentelemetry/github.com/gorilla/hypermux/mux.go
tim-mwangi: ok.

PR #61, opened 2020-11-26
Propagation support

jcchavezs:
This PR adds support for propagation config.


config/config.pbloader.go
```diff
@@ -24,6 +24,27 @@ func (x *AgentConfig) loadFromEnv(prefix string, defaultValues *AgentConfig) {
 		x.DataCapture = new(DataCapture)
 	}
 	x.DataCapture.loadFromEnv(prefix+"DATA_CAPTURE_", defaultValues.DataCapture)
+	if rawVals, ok := getArrayStringEnv(prefix + "PROPAGATION_FORMATS"); ok {
+
```
mohit-a21, line 28: nit: Extra line.

config/config.pbloader.go
```diff
@@ -24,6 +24,27 @@ func (x *AgentConfig) loadFromEnv(prefix string, defaultValues *AgentConfig) {
...
+		vals := []PropagationFormat{}
+		for _, rawVal := range rawVals {
+			vals = append(vals, PropagationFormat(PropagationFormat_value[rawVal]))
+		}
+		x.PropagationFormats = vals
+	} else if len(rawVals) == 0 && len(defaultValues.PropagationFormats) != 0 {
```
mohit-a21, line 34: Is there any case i am missing if we move `len(rawVals) == 0` to above  to  be like `rawVals, ok := getArrayStringEnv(prefix + "PROPAGATION_FORMATS"); ok && len(rawVals) > 0`. It seems like a redundant check here as part of else if.

mohit-a21 (approved):

config/config.pbloader.go
jcchavezs: Actually this check is redundant so I am removing it. Good catch.

PR #69, opened 2020-12-15
chore: adds release process.

jcchavezs:
This PR adds the release script for the Go agent.

Ping @mohit-a21 @davexroth @pavolloffay


mohit-a21:
@jcchavezs Can you outline the release workflow? In what case, one will have to run the script?

README.md
```diff
@@ -169,3 +169,7 @@ for unit tests only
 ```bash
 make test-unit
 ```
+
+### Releasing
+
+Run `./release.sh <version_number>`.
```
pavolloffay, line 175: what is the release number? An example would be helpful

README.md
```diff
@@ -169,3 +169,7 @@ for unit tests only
 ```bash
 make test-unit
 ```
+
+### Releasing
```
pavolloffay, line 173: Could you please explain by words what the release script does?

release.sh
```diff
@@ -0,0 +1,85 @@
...
+git tag -a "$VERSION" -m "Version $VERSION"
+
+NEW_VERSION="$MAJOR.$MINOR.$(($PATCH+1))-dev"
+
+write_version_file $NEW_VERSION $VERSION_FILE
+git add $VERSION_FILE
+
+git commit -m "chore(version): prepares for next version $NEW_VERSION"
+
+git push origin $MAIN_BRANCH || { rollback_changes $VERSION ; echo "Failed to push to $MAIN_BRANCH" ; exit 1 }
```
pavolloffay, line 83: Many people put origin as their for (including myself). Hence the should be configurable or use the full repo path (forked repos usually configure release separately anyways)

release.sh
```diff
@@ -0,0 +1,85 @@
...
+
+git push --tags
```
pavolloffay, line 85: missing remote?

README.md
jcchavezs: Sure.

release.sh
jcchavezs: As of now, only maintainers can do a release (and none of us need a fork) and in that case I would keep this script simple: "convention over configuration"

release.sh
jcchavezs: Not sure what you mean, do you think this isn't enough?

release.sh
pavolloffay: I use always forks!!!!

release.sh
pavolloffay: To which remote does the push pushes tags? I use always forks and use origin for my fork.

release.sh
pavolloffay: Keep the upstream clean ;)

release.sh
jcchavezs: As I said, I rather keep this simple and asume the releaser uses `origin` (in practice it is going to be me as of now). Later we can solve the fork thing.

PR #71, opened 2020-12-16
Fix content type lookup

jcchavezs:
Userland code can set multiheaders directly (separated by `;`) instead of using the Header.Add API. This requires to inspect the headers using contains rather than assert equal values.

Ping @davexroth

### DIFF
--- a/instrumentation/opentelemetry/github.com/gorilla/hypermux/mux_test.go
+++ b/instrumentation/opentelemetry/github.com/gorilla/hypermux/mux_test.go
@@ -28,12 +28,13 @@ func findAvailablePort() (int, error) {
 	return 0, errors.New("failed to find an available port")
 }
 
-func TestSpanNameIsSetCorrectly(t *testing.T) {
+func TestSpanRecordedCorrectly(t *testing.T) {
 	_, flusher := internal.InitTracer()
 
 	r := mux.NewRouter()
 	r.HandleFunc("/things/{thing_id}", func(rw http.ResponseWriter, r *http.Request) {
-		rw.Header().Set("content-type", "application/json")
+		rw.Header().Add("content-type", "application/json")
+		rw.Header().Add("content-type", "charset=utf-8")
 		rw.Header().Set("request_id", "xyz123abc")
 		rw.WriteHeader(202)
 		rw.Write([]byte(`{"id":123}`))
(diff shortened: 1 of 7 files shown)


instrumentation/opentelemetry/github.com/gorilla/hypermux/mux_test.go
```diff
@@ -28,12 +28,13 @@ func findAvailablePort() (int, error) {
 	return 0, errors.New("failed to find an available port")
 }
 
-func TestSpanNameIsSetCorrectly(t *testing.T) {
+func TestSpanRecordedCorrectly(t *testing.T) {
```
jcchavezs, line 31: Not important

sdk/net/http/contenttype_go1.13.go
```diff
@@ -0,0 +1,35 @@
...
+	"net/http"
+	"strings"
+)
+
+// shouldRecordBodyOfContentType checks if the body is meant
+// to be recorded based on the content-type and the fact that body is
+// not streamed.
+func shouldRecordBodyOfContentType(h http.Header) bool {
+	for _, contentTypeAllowed := range contentTypeAllowList {
+		lContentType := strings.ToLower(h.Get("content-type"))
```
jcchavezs, line 15: In go 1.13 we can only access the first value of a header so if the header is `"charset=utf-8" and  `"application/json"` we can't detect it. This is the best we can do for go 1.13.

sdk/net/http/contenttype_go1.14.go
```diff
@@ -0,0 +1,39 @@
...
+	"net/http"
+	"strings"
+)
+
+// shouldRecordBodyOfContentType checks if the body is meant
+// to be recorded based on the content-type and the fact that body is
+// not streamed.
+func shouldRecordBodyOfContentType(h http.Header) bool {
+	for _, contentTypeAllowed := range contentTypeAllowList {
+		for _, contentType := range h.Values("content-type") {
```
jcchavezs, line 17: In go 1.14 we can access all the values fro a header, hence we iterate over them and look for the desired content-type.

sdk/net/http/contenttype_test.go
```diff
@@ -7,20 +7,43 @@ import (
...
+func TestKeyLookupSuccessOnSet(t *testing.T) {
 	tCases := []struct {
 		contentType  string
 		shouldRecord bool
 	}{
 		{"text/plain", false},
 		{"application/json", true},
 		{"Application/JSON", true},
+		{"application/json", true},
+		{"application/json; charset=utf-8", true},
```
jcchavezs, line 19: This test makes sure we don't have a regression in the future.

instrumentation/opentelemetry/github.com/gorilla/hypermux/mux_test.go
```diff
@@ -74,10 +75,13 @@ func TestSpanNameIsSetCorrectly(t *testing.T) {
 	assert.Equal(t, "/things/{thing_id}", span.Name)
 	assert.Equal(t, span.SpanKind, trace.SpanKindServer)
 
+	fmt.Println(span.Attributes)
```
jcchavezs, line 78:
```suggestion
```

jcchavezs: chore: removes unnecessary code.

davexroth (approved):

PR #77, opened 2021-01-11
Update agent-config to latest

mohit-a21:
Latest agent config uses snake casing instead of camel
casing.


jcchavezs:
@mohit-a21 do you mind running `go mod tidy`?

PR #80, opened 2021-02-03
Expose shouldRecordBodyOfContentType for use by other packages

rnburn:
This will allow our ext-authz service to reuse the should-record logic.

Note envoy's protocol uses a [different format](https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/attribute_context.proto#service-auth-v3-attributecontext-httprequest) for representing headers; hence, the addition of the HeaderAccessor interface.

### DIFF
--- a/sdk/net/http/contenttype.go
+++ b/sdk/net/http/contenttype.go
@@ -1,7 +1,6 @@
 package http
 
 import (
-	"net/http"
 	"strings"
 )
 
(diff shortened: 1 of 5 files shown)


rnburn:
@davexroth

jcchavezs (approved):
LGTM

PR #82, opened 2021-02-09
Expose SetAttributesFromHeaders

rnburn:
Allows other packages to reuse the logic for setting header attributes. (Will be used by an authz service).


rnburn:
@davexroth @jcchavezs - could you take a look please? thanks.

davexroth (approved):

PR #83, opened 2021-02-12
Adds Filter interface to SDK.

ryanericson:
- Filter takes map of request span attributes and returns decision
whether request should be filtered/blocked. Filter may add
attributes to span.
- Added filtering capability to http instrumentation.


ryanericson:
@jcchavezs, @davexroth tests, examples, and GRPC changes are TODO. But I wanted to get feedback first if this pattern looks good.

jcchavezs:
Hi @ryanericson thanks for working this out. I have a few concerns about the implementation, the main one being the need to look at the span attributes. [As far as I understood from the spec](https://github.com/hypertrace/specification/blob/main/agent/filtering.md#evaluate-request-headers), we look at headers/body from the request, I don't think we need to wait for this values to be span attributes and then check the span attributes. This is mostly because we are not suppose to look at span attributes on instrumentation libraries.

My other main concern is usability. Userland code will look like `hyperhttp.WrapHandler(sdkhttp.Options{})` which is importing two packages for a single functionality. More so, if passing a filter: `hyperhttp.WrapHandler(sdkhttp.Options{BlockingFilter: sdkhttp.MyFilter})`. In that case I'd use functional arguments `hyperhttp.WrapHandler(hyperhttp.WithBlockingFilter(hyperhttp.MyFilter))` that internally feed the `sdkhttp.Options`.

ryanericson:
> Hi @ryanericson thanks for working this out. I have a few concerns about the implementation, the main one being the need to look at the span attributes. [As far as I understood from the spec](https://github.com/hypertrace/specification/blob/main/agent/filtering.md#evaluate-request-headers), we look at headers/body from the request, I don't think we need to wait for this values to be span attributes and then check the span attributes. This is mostly because we are not suppose to look at span attributes on instrumentation libraries.
> 

Thanks for the fast feedback @jcchavezs. I didn't know about the specification. I see that in there the filter just looks at request headers and body, but filtering needs to be done on entire of request properties. This will allow us to support all kinds of filtering scenarios, such as filtering based on query/GET params, filtering based on HTTP method (POST, PATCH etc.), filtering based on HTTP protocol versions, and so on. So I think we the filtering method should just be just one method that looks at entire request. To represent an entire request we can either have another abstraction of request just for Hypertrace, or we can represent it with a bag of request span attributes. Considering that this is for a tracing agent I don't think it's bad to use span attributes representation.  

> My other main concern is usability. Userland code will look like `hyperhttp.WrapHandler(sdkhttp.Options{})` which is importing two packages for a single functionality. More so, if passing a filter: `hyperhttp.WrapHandler(sdkhttp.Options{BlockingFilter: sdkhttp.MyFilter})`. In that case I'd use functional arguments `hyperhttp.WrapHandler(hyperhttp.WithBlockingFilter(hyperhttp.MyFilter))` that internally feed the `sdkhttp.Options` or create `hyperhttp.Options` for every server package.

Makes sense about hiding `sdkhttp.Options` I can create an Options type for each server package as you suggests.

jcchavezs:
Thanks for the response @ryanericson 

> Thanks for the fast feedback @jcchavezs. I didn't know about the specification. I see that in there the filter just looks at request headers and body, but filtering needs to be done on entire of request properties. This will allow us to support all kinds of filtering scenarios, such as filtering based on query/GET params, filtering based on HTTP method (POST, PATCH etc.), filtering based on HTTP protocol versions, and so on.

I suggest you open an issue on the spec to discuss this. The spec was introduced early and mostly related to the work on java so this new implementation might make it evolve. In the other hand I suggest you come up with use cases for e.g. blocking by HTTP method or HTTP protocol (which I am not against to as long as we actually have use cases for it).

> So I think we the filtering method should just be just one method that looks at entire request. To represent an entire request we can either have another abstraction of request just for Hypertrace, or we can represent it with a bag of request span attributes. Considering that this is for a tracing agent I don't think it's bad to use span attributes representation.

I see many many disadvantages on the usage of span attributes for this. First being that we have to do [this kind of things](https://github.com/hypertrace/goagent/pull/83/files#diff-5f15e25537f4b9908f4e8c90696366146a3bb4594020e096567f2fa0c5af9411R11) to keep track of the attributes being added, on the other hand there is configurable data capture which might/might not let you capture the request/response body. That is why I'd rather use the request data.

Based on the work I'd done in [zipkin-go](https://github.com/openzipkin/zipkin-go/blob/f113cc090cee63e59d38cdda0dbeea9cad51ebdb/middleware/http/transport.go#L221) I pass the values to the user instead of allowing them access the full request. Reason for this is that we should avoid leaks for body and ensure the hand-of of the body to the wrapped handler.

Here you can find a example implementation: https://gist.github.com/jcchavezs/f83ddd1adebbd2d31d321ce2afa5a051

My questions then would be (cc @davexroth). Are there use cases for blocking by method? if so, is this blocking logic in conjunction with other elements like headers or query params? Only thing that is clear to me is that `FilterOutByBody` method should be separated but all the others can be either separated or coalesced.

jcchavezs:
> EDITED: @jcchavezs. However the Filter interface is in sdk package right? If someone were to use our http instrumentation with their own filtering logic they would need to import sdk package. We're ok with that? Or are we talking about going to the extent of defining a copy of the Filter interface in hyperhttp to completely hide SDK?

I think this repo is mostly including the interface only. In my proposal above I suggest using a struct and not an interface. In such case I am OK using the sdk `http.ServerRequestFilter`.

sdk/instrumentation/net/http/handler.go
```diff
@@ -45,16 +57,39 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		span.SetAttribute(key, value)
 	}
 
-	span.SetAttribute("http.url", r.URL.String())
+	url := r.URL.String()
+	span.SetAttribute("http.url", url)
+
+	// run url filters
+	for _, urlEvaluator := range h.urlFilters {
```
jcchavezs, line 64: I think all the filters should go up as in this line https://github.com/hypertrace/goagent/pull/83/files#diff-c9de109efb7aa8920966082d6c5481c3324292b143637ff5c5f3f2174a5e1b9dR52 we are passing the request to the userland handler.

sdk/instrumentation/net/http/handler.go
```diff
@@ -17,21 +19,31 @@ type handler struct {
 	defaultAttributes        map[string]string
 	spanFromContextRetriever sdk.SpanFromContext
 	dataCaptureConfig        *config.DataCapture
+	urlFilters               []filter.URLFilter
+	headersFilters           []filter.HeadersFilter
+	bodyFilters              []filter.BodyFilter
+}
+
+// Options for HTTP handler instrumentation
+type Options struct {
```
jcchavezs, line 28: I think it is better to put the filters inside a single field in options as options other options might come and then it will look very varied.

sdk/filter/filter.go
```diff
@@ -0,0 +1,10 @@
+package filter
```
jcchavezs, line 1: Current types semantics seem ambiguous to me, does `block` mean `true` or `block` means `false`, could we use imperatives?

sdk/instrumentation/net/http/handler.go
ryanericson: Makes sense.

sdk/instrumentation/net/http/handler.go
ryanericson: The link doesn't take me anywhere for some reason. Can you elaborate what you mean by going up?

sdk/filter/filter.go
ryanericson: I was chatting with @davexroth and there was a brief mention of wanting to keep the filter generic as in the future we will want to use it for more than just blocking.
Ah, the span should be there as argument, let me add that.

jcchavezs: These types return true or false so I don't think we can do more than block/don't block. If we want to add other kind of functionality we better add other types or interfaces. Imagine you want to cover other use cases than blocking (would be nice to know some), how would you differentiate those from the blocking to not block?

I truly think we should be explicit rather than cryptic as this is going to impact user workflow.

cc @davexroth

sdk/instrumentation/net/http/handler.go
davexroth, line 64: All this filter logic should be encapsulated in one filter package which hides all the details.  Then each client (instrumentation lib) only needs to call one function rather than iterating over each filter, ie
`
if filter.evaluateURL(span, url) {
  w.WriteHeader(http.StatusForbidden)
  return
}
`

I think this is where you were also going right JC?

sdk/instrumentation/net/http/handler.go
```diff
@@ -45,16 +57,39 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
-	span.SetAttribute("http.url", r.URL.String())
+	url := r.URL.String()
+	span.SetAttribute("http.url", url)
+
+	// run url filters
+	for _, urlEvaluator := range h.urlFilters {
+		if urlEvaluator(url) {
+			w.WriteHeader(http.StatusForbidden)
+			w.Write([]byte(fmt.Sprintf("403 - Blocked by URL filter")))
```
davexroth, line 67: do we also write the body in the other agents when we block?

sdk/filter/filter.go
davexroth: Agreed normally we should have better return types, but in this case I'm ok with just returning true/false, as this is what's being implemented in the other environments and how our blocking implementation has been across the languages.

As Ryan mentioned, I do want to keep this interface consistent across languages - as the dream is to be able to have this one common interface across agents, and then we'd have a single webasam filter which could be loaded in any of the runtimes.  When the interface deviates between implementations, it makes that more difficult.

sdk/instrumentation/net/http/handler.go
ryanericson: In envoy we do. I'll just get rid of the body.

sdk/instrumentation/net/http/handler.go
ryanericson: I see, done.

sdk/filter/filter.go
```diff
@@ -0,0 +1,44 @@
...
+
+import "github.com/hypertrace/goagent/sdk"
+
+// Filter runs on request evaluates whether request should be
+// blocked
+type Filter struct {
```
davexroth, line 7: how are all these filters going to be installed?  I was thinking we could mirror https://github.com/hypertrace/javaagent/blob/main/filter-api/src/main/java/org/hypertrace/agent/filter/MultiFilter.java so we'd have something like:

`
type MultiFilter {
  filters []filter
}

NewMultiFilter(filters []filter) *MultiFilter {
...
}


type filter interface {
  evaluateRequestURL        func(span sdk.Span, URL string) bool
  evaluateRequestHeaders func(span sdk.Span, headers map[string][]string) bool
  evaluateRequestBody       func(span sdk.Span, body []byte) bool
}
`

ryanericson: It will look like what's in `handler_test.go`. I agree this is nicer, but I'll keep filter as struct so that we can easily tell if the filter cares about body or not, and can easily avoid body extraction if it doesn't and body capture is not configured. (@jcchavezs's idea).

davexroth: I think that may be overkill. For the cases where there's a filter configured, we can assume we're always going to have body capture enabled.

Even if the particular filter isn't making use of the body, there's no harm in it having the implement `evaluateRequestBody` but have it a noop.

sdk/instrumentation/net/http/handler.go
```diff
@@ -17,21 +18,27 @@ type handler struct {
 	defaultAttributes        map[string]string
 	spanFromContextRetriever sdk.SpanFromContext
 	dataCaptureConfig        *config.DataCapture
+	requestFilter            filter.Filter
```
davexroth, line 21: should this just be called `filter`, as the same filter will potentially handle both request and response data.

sdk/instrumentation/net/http/handler.go
```diff
@@ -45,16 +52,33 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	// run filters on headers
+	if h.requestFilter.EvaluateHeaders(span, headers) {
+		w.WriteHeader(http.StatusForbidden)
+		return
+	}
+
+	shouldRecordBody := h.dataCaptureConfig.HttpBody.Request.Value && ShouldRecordBodyOfContentType(headerMapAccessor{r.Header})
+	shouldFilterByBody := len(h.requestFilter.BodyEvaluators) > 0
```
davexroth, line 77: related to the other comment, I think this check may be un-necessary. If we're capturing the body, we should *always* call `filter.evaluate[Request|Response]Body`, even if it results in a noop.

sdk/filter/filter.go
ryanericson: I'm ok with tying it with body data capture.

sdk/instrumentation/net/http/handler.go
ryanericson: done

sdk/instrumentation/net/http/handler.go
ryanericson: done

sdk/filter/filter.go
```diff
@@ -2,43 +2,68 @@ package filter
...
-
-// EvaluateURL runs URL evaluators in the filter
-func (f Filter) EvaluateURL(span sdk.Span, url string) bool {
-	for _, urlEvaluator := range f.URLEvaluators {
-		if urlEvaluator(span, url) {
+// Filter evaluates whether request should be blocked
+type Filter interface {
+	EvaluateURL(span sdk.Span, url string) bool
+	EvaluateHeaders(span sdk.Span, headers map[string][]string) bool
+	EvaluateBody(span sdk.Span, body []byte) bool
```
jcchavezs, line 9: I still believe we should use struct with null checks rather than an interface, otherwise we will read the body before the userland code no matter the user intent.

Can't we do 

```go
type Filter struct {
    EvaluateURL func (span sdk.Span, url string) bool
    EvaluateHeaders func (span sdk.Span, headers map[string][]string) bool
    EvaluateBody func (span sdk.Span, body []byte) bool
}
```

In addition, consider using the headerAccessor for headers as that is what we use for HTTP (cc https://github.com/hypertrace/goagent/pull/87)

sdk/instrumentation/net/http/handler.go
```diff
@@ -87,12 +88,12 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		// Only records the body if it is not empty and the content type
 		// header is not streamable
-		if shouldRecordBody && len(body) > 0 {
+		if len(body) > 0 {
 			span.SetAttribute("http.request.body", string(body))
 		}
 
 		// run body filters
-		if h.requestFilter.EvaluateBody(span, body) {
+		if h.filter.EvaluateBody(span, body) {
```
jcchavezs, line 96: Notice this filter is only applied if the dataCapture for body is enabled which is IMO an ortogonal concern to recording of body.

jcchavezs:
@ryanericson I left a couple of comments, other than those it LGTM.

sdk/filter/filter.go
```diff
@@ -0,0 +1,69 @@
+package filter
+
+import "github.com/hypertrace/goagent/sdk"
+
+// Filter evaluates whether request should be blocked
```
jcchavezs, line 5:
```suggestion
// Filter evaluates whether request should be blocked, `true` blocks the request and `false` continues it.
```

ryanericson: done

sdk/instrumentation/net/http/handler.go
ryanericson: See comment here: https://github.com/hypertrace/goagent/pull/83#discussion_r578867109.

sdk/filter/filter.go
ryanericson: I think we want consistency in interface with Java Agent's. And we're ok with tying body filtering with data capture. See here: https://github.com/hypertrace/goagent/pull/83#discussion_r578839469.

Right, I looked at headerAccessor but that's specific to HTTP instrumentation, whereas we want filter to be generic for other instrumentations e.g. grpc.

sdk/filter/README.md
```diff
@@ -0,0 +1,24 @@
...
+Filtering capability as specified in [Hypertrace filter specification](https://github.com/hypertrace/specification/blob/main/agent/filtering.md) can be added to instrumentations.
+
+Example
+```go
+// Filter requests containing "foo" in URL
+type FooURLFilter struct {
+}
+
+func (FooURLFilter) EvaluateURL(span sdk.Span, url string) bool {
+	if strings.Contains(url, "foo") {
```
jcchavezs, line 11: I think it is easier to just return `strings.Contains(url, "foo")`

sdk/filter/filter.go
jcchavezs: I really like consistency but I also believe we should stick to contracts and not implementations. I think every language has its own trade-offs and implementation strategies. As of Go I think the right approach is using a struct and we can keep consistency on the contract, we will probably do the same in languages that don't have interfaces (e.g. javascript, Ruby). Personally (as I told @davexroth) I'd prefer instrumentation to have the less impact on the user code and reading the body for nothing does not sound the best approach in Go (maybe in other languages where you don't need to reassign the body every time you read it). On the other hand I think we have spent too much time on this back and forth so I'd let you guys decide.

sdk/filter/README.md
ryanericson: right, done.

sdk/filter/filter.go
davexroth: I think I'm missing the case where we're reading the body un-necessarily (which is bad!). can you expand on that case?

jcchavezs: Users don't necessarily read the body, e.g. unsupported content-type (there are other cases but let's assume they all happen in an upper layer, e.g. authorization). In our case we will anyways read the body on behalf of the user and the user will probably dismiss the request without making use of the body content. By doing that we are impacting the user workflow and it can be more intensive if we talk about big payloads.

jcchavezs:
As discussed with @davexroth there won't be filtering applied to body if data capture isn't enabled.

PR #92, opened 2021-02-24
chore(config): fixes proto version in the code generation for config.

jcchavezs:
Ping @pavolloffay


pavolloffay (approved):
I would prefer that the dockerfile would download and use a specific version of tooling rather than relying on the developer's host system

jcchavezs:
You can create an issue for that.

PR #94, opened 2021-03-04
feat: Init tracer with resources

mohit-a21:
Extended hypertrace Init with a new API InitWithResources
which defines immutable resources in the case of otel tracer

### Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+.idea/**
+.vscode/**
+instrumentation/hypertrace/google.golang.org/hypergrpc/examples/.DS_Store
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ func main() {
 
 Config values can be declared in config file, env variables or code. For further information about config check [this section](config/README.md).
 
+Immutable resource information for the env can be added by calling `hypertrace.InitWithResources(cfg)`
+
 ## Package net/hyperhttp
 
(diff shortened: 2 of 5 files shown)


jcchavezs:
I think I mentioned it this in the past but I believe resources should be
part of the agent config because that feature will be shared across agents
and also it's a property of the agent itself. Does that make sense?

mohit-a21:
> I think I mentioned it this in the past but I believe resources should be part of the agent config because that feature will be shared across agents and also it's a property of the agent itself. Does that make sense? cc @pavolloffay

@jcchavezs I agree with you that we should be able to define resources from agent-config. But I think, there can be cases where some resources are evaluated at the time of application run and are dynamic in nature. We should be able to pass those resources separately.

pavolloffay:
I think there should be a programmatic API to set the resources. How does the workflow work in the goagent? Can a developer pass a config instance to factory method and get configured tracer/agent with the resources? If so then we should take this route. However if the tracer/agent can be created and configured without supplying a config then exposing an API to set resources if totally fine.

jcchavezs:
>  But I think, there can be cases where some resources are evaluated at the time of application run and are dynamic in nature. We should be able to pass those resources separately.

That is fine, still you can pass that info over config like in this example where we pass the service name: https://github.com/hypertrace/goagent#getting-started

PR #101, opened 2021-03-05
feat: adds exporters support.

jcchavezs:
## Description
This PR adds support for exporters in the opentelemetry tracer.


instrumentation/opentelemetry/init.go
```diff
@@ -45,16 +48,36 @@ func makePropagator(formats []config.PropagationFormat) propagation.TextMapPropa
...
+
+		if !cfg.GetReporting().GetSecure().GetValue() {
+			opts = append(opts, otlpgrpc.WithInsecure())
+		}
+
+		batcherExporter, err = otlp.NewExporter(
+			context.Background(),
+			otlpgrpc.NewDriver(opts...),
+		)
+	} else {
```
jcchavezs, line 69:
```suggestion
	} else { // Zipkin is the default exporter
```

PR #102, opened 2021-03-07
configured propagation formats ignored

davexroth:
## Description
If the propagation formats were defined in the config file, there were getting overwritten by the default value.

### Checklist:
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -57,23 +57,31 @@ func TestSnakeYAMLLoadSuccess(t *testing.T) {
 
 func TestConfigLoadFromEnvOverridesWithEnv(t *testing.T) {
 	cfg := &AgentConfig{
-		ServiceName: String("my_service"),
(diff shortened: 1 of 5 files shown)


config/config_test.go
```diff
@@ -7,15 +7,22 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+// A number of these tests set environment variables.
+// When setting an env var, ensure that it is unset
+// at the end of the test so as not to impact other tests
+
```
jcchavezs, line 13: @davexroth indeed! We might need to do further refactoring if at some point we want to run this tests in parallel (which we should).

jcchavezs (approved):

PR #103, opened 2021-03-10
feat: Add support for multiple tracer providers

mohit-a21:
Added InitService which internally creates a new TracerProvider
used for creating services with a different set of resource attributes
then the default service.

### DIFF
--- a/instrumentation/opentelemetry/init.go
+++ b/instrumentation/opentelemetry/init.go
@@ -31,8 +31,8 @@ var batchTimeout = time.Duration(200) * time.Millisecond
 
 var (
 	traceProviders map[string]*sdktrace.TracerProvider
-	batcher        *sdktrace.BatchSpanProcessor
-	sampler        sdktrace.Sampler
+	globalBatcher  *sdktrace.BatchSpanProcessor
+	globalSampler  sdktrace.Sampler
 	initialized    = false
 	mu             sync.Mutex
 )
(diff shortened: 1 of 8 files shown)


instrumentation/opentelemetry/init.go
```diff
@@ -59,37 +77,80 @@ func Init(cfg *config.AgentConfig) func() {
 		log.Fatal(err)
 	}
 
+	defaultBatcher := sdktrace.NewBatchSpanProcessor(zipkinExporter)
```
jcchavezs, line 80: I would not call it defaultBatcher as it is the only batcher. How about just calling it batcher?

instrumentation/opentelemetry/init.go
```diff
@@ -59,37 +77,80 @@ func Init(cfg *config.AgentConfig) func() {
...
+	traceProviders = make(map[string]*sdktrace.TracerProvider)
+	batcher = defaultBatcher
+	sampler = defaultSampler
+	initialized = true
 	return func() {
+		mu.Lock()
+		defer mu.Unlock()
+		batcher.Shutdown(context.Background())
+		for _, tracerProvider := range traceProviders {
+			tracerProvider.Shutdown(context.Background())
```
jcchavezs, line 108: I think you can also delete the traceprovider from the map.

mohit-a21: Done.

instrumentation/opentelemetry/init.go
mohit-a21: Done. Changed the global batcher to `globalBatcher`.

jcchavezs (approved):

jcchavezs:
Great work @mohit-a21!

PR #105, opened 2021-03-17
Add EvaluateURLAndHeaders method to Filter interface.

ryanericson:


davexroth (approved):

jcchavezs:
If we add this functionality, are we still in time to drop `EvaluateURL` and `EvaluateHeaders`? I think it is going to be much easier to have one rather than have three methods as `EvaluateURLAndHeaders` supersedes `EvaluateURL` and `EvaluateHeaders`. It is also less confusing as there might be misconceptions about which executes first.

cc @davexroth

jcchavezs:
@ryanericson could you please elaborate on why this is needed and what use case are we trying to achieve because independent evaluation of headers and URL can be done anyways?

jcchavezs:
If we add this functionality, are we still in time to drop EvaluateURL and EvaluateHeaders? I think it is going to be much easier to have one rather than have three methods as EvaluateURLAndHeaders supersedes EvaluateURL and EvaluateHeaders. It is also less confusing as there might be misconceptions about which executes first.

cc @davexroth

davexroth:
We needed to do this because crossing language boundaries is expensive in some cases, and when we combined the call saw an improvement in our benchmarks.

We had kept the original single EvaluateURL and EvaluateHeaders in cases where you only wish to (or can) evaluate url or headers.

I could be persuaded to remove them, as you can always call EvaluateURLAndHeaders with either an empty url or empty header map, although it's debatable whether that or individual methods is more confusing.

jcchavezs:
> I could be persuaded to remove them, as you can always call EvaluateURLAndHeaders with either an empty url or empty header map, although it's debatable whether that or individual methods is more confusing.

I think with individual methods along with join method there is always room for confusion. Joint method along may or may not be confusing.

PR #112, opened 2021-04-06
Default to otlp exporter

shashank11p:
Depends on https://github.com/hypertrace/agent-config/pull/59


jcchavezs:
This is a breaking change, do we need to introduce this?

davexroth:
> This is a breaking change, do we need to introduce this?

this is a change we'll need to make at some stage, so I'd prefer to do it sooner rather than later.

jcchavezs:
CI is failing because some tests assumed the reporter was zipkin.

shashank11p:
> CI is failing because some tests assumed the reporter was zipkin. I am not sure how we should prevent this in the future as the error is very cryptic:
> 
> ```
> 2021/04/08 10:58:05 rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp: address http://127.0.0.1:45253: too many colons in address"
> ```
> 
> Can agents do a best effort to validate the provided endpoint is acceptable or just let the user fail on this? cc @pavolloffay @davexroth

Thanks @jcchavezs for fixing the test. This error is because in goagent, otlp exporter expects only `ip:port` as endpoint, while zipkin exporter expects `http://ip:port`. We can document this if not validating in agent

PR #120, opened 2021-05-19
feat: adds readback span support for using it in the filter interface.

jcchavezs:
## Description

This PR adds support for ReadbackSpan which might give access to the attributes in filtering.

Ping @ryanericson @davexroth

### DIFF
--- a/instrumentation/opentelemetry/readbackspan.go
+++ b/instrumentation/opentelemetry/readbackspan.go
@@ -1,0 +1,44 @@
+package opentelemetry
+
+import (
+	"context"
+	"sync"
+
+	"github.com/hypertrace/goagent/sdk"
+	"go.opentelemetry.io/otel"
+)
+
+var _ sdk.ReadbackSpan = (*ReadbackSpan)(nil)
+
+type ReadbackSpan struct {
+	sdk.Span
+	readAttrs *map[string]interface{}
+	m         sync.Mutex
+}
+
+func (rs *ReadbackSpan) GetAttribute(key string) interface{} {
+	if v, ok := (*rs.readAttrs)[key]; ok {
+		return v
+	}
+
+	return nil
+}
+
+func (rs *ReadbackSpan) SetAttribute(key string, value interface{}) {
+	rs.m.Lock()
+	defer rs.m.Unlock()
+
+	(*rs.readAttrs)[key] = value
+	rs.Span.SetAttribute(key, value)
+}
+
+func startReadbackSpan(provider getTracerProvider) sdk.StartReadbackSpan {
+	ss := startSpan(provider)
+	return func(ctx context.Context, name string, opts *sdk.SpanOptions) (context.Context, sdk.ReadbackSpan, func()) {
+		ctx, s, ender := ss(ctx, name, opts)
+		readAttrs := map[string]interface{}{}
+		return context.WithValue(ctx, &readAttrsKey, readAttrs), &ReadbackSpan{s, &readAttrs, sync.Mutex{}}, ender
+	}
+}
+
+var StartReadbackSpan = startReadbackSpan(otel.GetTracerProvider)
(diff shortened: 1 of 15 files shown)


sdk/readbackspan.go
```diff
@@ -0,0 +1,19 @@
+package sdk
+
+import "context"
+
+// ReadbackSpan is a span whose attributes can be read. It is important
+// to bear in mind that only those attributes being recorded through this
+// interface will be available for read. This is, an existing span where
+// attributes have been already recorded can be turned into a ReadbackSpan
+// but the attributes won't be available.
+type ReadbackSpan interface {
```
ryanericson, line 10: Looks reasonable to me. It means in instrumentation where we need to support filtering then we should use `ReadbackSpan` instead of `Span`? What are our thoughts of just making `Span` have readable attributes?

sdk/readbackspan.go
```diff
@@ -0,0 +1,19 @@
...
+	Span
+	GetAttribute(key string) interface{}
```
ryanericson, line 12: Also need a method to get all attributes.

instrumentation/opentelemetry/readbackspan.go
```diff
@@ -0,0 +1,44 @@
...
+var _ sdk.ReadbackSpan = (*ReadbackSpan)(nil)
+
+type ReadbackSpan struct {
+	sdk.Span
+	readAttrs *map[string]interface{}
+	m         sync.Mutex
+}
+
+func (rs *ReadbackSpan) GetAttribute(key string) interface{} {
+	if v, ok := (*rs.readAttrs)[key]; ok {
```
ryanericson, line 20: We'll need to take the lock here too, right?

sdk/readbackspan.go
jcchavezs: I am curious about the reason, this will be called inside the filter and I assume filter knows what attributes to read, isn't?

tyroneyu: We moved that further into libtraceable so that we didn't have to keep updating all the libtraceable users (ext_cap, nginx, envoy, etc) if the required attributes change. At the moment, we would send all span attributes.

jcchavezs: Then I would change this signature into return all attributes.

ryanericson:
@jcchavezs as this is blocking @tyroneyu can you please prioritize  review when you get a chance so that he's unblocked on this filter's work.

ryanericson (approved):

instrumentation/opentelemetry/readbackspan_test.go
```diff
@@ -0,0 +1,31 @@
...
+	s.SetAttribute("test_key_4", "abc")
+	err := errors.New("xyz")
+	s.SetAttribute("test_key_4", err)
+	attributes := s.GetAttributes()
+	assert.Equal(t, attributes["test_key_1"].(bool), true)
+	assert.Equal(t, attributes["test_key_2"].(int64), int64(1))
+	assert.Equal(t, attributes["test_key_3"].(float64), 1.2)
+	assert.Equal(t, attributes["test_key_4"].(error), err)
+
+	assert.Equal(t, fmt.Sprintf("%v", attributes["test_key_1"].(bool)), "true")
```
jcchavezs, line 27: I am curious about these assertions, why exactly we do this given we have assertions on top?

tyroneyu: I duplicated it thinking it didn't hurt.  This is most likely due to my Go inexperience and just learning about interface{}'s .  From the consuming end of GetAttributes(), we always use a string to pass into libtraceable. 

I'll remove the second block of asserts.

sdk/filter/filter.go
```diff
@@ -9,4 +9,8 @@ type Filter interface {
 
 	// EvaluateBody can be used to evaluate the body content
 	EvaluateBody(span sdk.Span, body []byte) bool
+
+	// Evaluate can be used to evaluate the span attributes. Notice span will contain at most
+	// pre-processing attributes (e.g. request attributes in a HTTP/RPC call).
+	Evaluate(span sdk.ReadbackSpan) bool
```
jcchavezs, line 15: This method allows to filter based on the span attributes. The reason why we keep this method along with the others is because the easiest cases are to evaluate by header, url and body whereas the span attributes requires the instrumentation to create a ReadbackSpan cc @tyroneyu @ryandens

tyroneyu: Regarding the existing two interface methods, we still have the problem with
https://traceableai.atlassian.net/wiki/spaces/Engineering/pages/1265139838/Blocking+rules+-+evaluation+order+of+precedence right?  Specifically when we allow an IP address since the required headers aren't being sent to either call.

PR #123, opened 2021-07-09
fix: Update http handler to record host

prodion23:
## Description
The go http package copies the http host header to `request.Host` field and deletes the header from the map.
This change allows us to record the http host.

https://github.com/golang/go/blob/ab4085ce84f8378b4ec2dfdbbc44c98cb92debe5/src/net/http/request.go#L154-L155

The above behavior prevents `SetAttributesFromHeaders` method from being able to record the host header.

### Testing
The new test asserts that `request.Host` is added to the spans attributes.

### Checklist:

### Documentation

### DIFF
--- a/sdk/instrumentation/net/http/handler.go
+++ b/sdk/instrumentation/net/http/handler.go
@@ -59,6 +59,9 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	url := r.URL.String()
 	span.SetAttribute("http.url", url)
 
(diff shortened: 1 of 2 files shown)


davexroth:
Can you confirm behavior with grpc requests?  Do we need a similar change to capture `rpc.request.metadata.:authority` ?

prodion23:
Looks like we do capture grpc `:authority` header - I think this issue was isolated to the go http package

davexroth (approved):

PR #124, opened 2021-07-12
fix: Always report rpc method & scheme

prodion23:
:method & :scheme are not reported - these fields are not present in the `metadata.FromIncomingContext(ctx)` metadata map.

## Description
method - grpc will error if a request is received that doesn't use POST, and the grpc spec defines post as the only valid http method, just apply a default for this attribute
scheme - peer.AuthInfo will be nil if no transport security is configured

### Testing
Edits to existing/new test cases for these attributes

### Checklist:

### DIFF
--- a/sdk/instrumentation/google.golang.org/grpc/server.go
+++ b/sdk/instrumentation/google.golang.org/grpc/server.go
@@ -12,6 +12,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/peer"
 	"google.golang.org/grpc/stats"
 	"google.golang.org/grpc/status"
 )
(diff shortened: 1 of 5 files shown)


davexroth (approved):

jcchavezs:
Could you please use http.MethodPost? Also is AuthInfo included implying
always https? Can't this be passed over http? What if AuthType returns
empty string?

prodion23:
@jcchavezs updated post 👍 

From what I understood, from the linked `peer.AuthInfo` comment, the AuthInfo field will only ever be present if communicating over https?
https://github.com/grpc/grpc-go/blob/master/peer/peer.go#L35-L37

sdk/instrumentation/google.golang.org/grpc/server.go
```diff
@@ -231,3 +237,14 @@ func WrapStatsHandler(delegate stats.Handler, spanFromContext sdk.SpanFromContex
 		dataCaptureConfig: internalconfig.GetConfig().GetDataCapture(),
 	}
 }
+
+func setSchemeAttributes(ctx context.Context, span sdk.Span) {
+	peer, ok := peer.FromContext(ctx)
+	if ok {
```
jcchavezs, line 243: If you do `!ok` this can be turned into early termination and we save one identation.

sdk/instrumentation/google.golang.org/grpc/server.go
```diff
@@ -231,3 +237,14 @@ func WrapStatsHandler(delegate stats.Handler, spanFromContext sdk.SpanFromContex
...
+		if peer.AuthInfo != nil {
```
jcchavezs, line 244: nit: I'd rather resolve the scheme on before hand and only call SetAttribute once.

jcchavezs:
> From what I understood, from the linked peer.AuthInfo comment, the AuthInfo field will only ever be present if communicating over https?
https://github.com/grpc/grpc-go/blob/master/peer/peer.go#L35-L37

Worth to add this line to the code explaining this. Remember to use a version instead of master whenever you link to hosted code because this could change and explanation could be lost.

jcchavezs:
Great work!

PR #125, opened 2021-07-13
Report v1 proto data

prodion23:
## Description
Applications that use an older protobuf format previously would not have their body captured. 

We need to use the `github.com/golang/protobuf/proto` proto instead because the conversion methods(`MessageV2`) are not available in the `google.golang.org/protobuf/proto`

### Testing
New test asserts we can capture v1 & v2 protobuf spans(v2 is what was previously present in the hello world sample app)

Ensured the sample app still reports body data, and tested a v1 protobuf application locally and it is now reporting that data.

### Checklist:


jcchavezs:
LGTM

PR #129, opened 2021-08-16
chore: relocate internal package used for testing.

jcchavezs:
## Description

This PR changes the internal package containing testing utils for tracing into a tracetesting package. This to make it explicit the testing purposes but also to organize the packages under internal.


prodion23 (approved):
Changes look good to me

PR #141, opened 2021-09-22
feat: adds span status support.

jcchavezs:
This PR adds support for status API in spans, allowing users to set the status of an operation despite the error value.

### DIFF
--- a/sdk/codes.go
+++ b/sdk/codes.go
@@ -1,0 +1,13 @@
+package sdk
+
+type Code uint32
+
+const (
+	// Unset is the default status code.
+	Unset Code = 0
+	// Error indicates the operation contains an error.
+	Error Code = 1
+	// Ok indicates operation has been validated by an Application developers
+	// or Operator to have completed successfully, or contain no error.
+	Ok Code = 2
+)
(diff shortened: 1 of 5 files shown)


sdk/codes.go
```diff
@@ -0,0 +1,13 @@
+package sdk
+
+type Code uint32
+
+const (
```
davexroth, line 5: https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto#L314

why not match with otel?

jcchavezs: What exactly do you mean? Values and names match, also this is copied from codes.Codes from otel-go.

davexroth: In the proto definition `STATUS_CODE_OK  = 1;`

jcchavezs: I based on https://github.com/open-telemetry/opentelemetry-go/blob/main/codes/codes.go#L23

jcchavezs: Changed.

davexroth (approved):

sdk/codes.go
tim-mwangi: I'd say we do what otel go does and not the proto, since that's what the otel agent is based on.

PR #146, opened 2021-09-29
feat: load tls CA certificate into the exporter client

tim-mwangi:
## Description
In order to be able to pass in a private root CA certificate to verify a server's certificate when exporting the spans.

### Testing
Added unit tests

### Checklist:
- [ ✅ ] I have added tests that prove my fix is effective or that my feature works

### Documentation
Since a new configuration has been added, this will need to be documented in our docs.


instrumentation/opentelemetry/init.go
```diff
@@ -101,6 +113,36 @@ func makeExporterFactory(cfg *config.AgentConfig) func() (sdktrace.SpanExporter,
...
+	tlsConfig.InsecureSkipVerify = !cfg.GetReporting().GetSecure().GetValue()
+	tlsConfig.RootCAs = createCaCertPoolFromFile(cfg)
+
+	return tlsConfig
+}
+
+// createCaCertPoolFromFile creates a CA Cert Pool from a file path containing
+// a raw CA certificate to verify a server certificate. The file path is the
+// reporting.cert_file config value.
+func createCaCertPoolFromFile(cfg *config.AgentConfig) *x509.CertPool {
```
jcchavezs, line 127: This function does not need to be aware of the whole `*config.AgentConfig`, we should just use `*config.CertFile`, the less the better.

instrumentation/opentelemetry/init.go
```diff
@@ -101,6 +113,36 @@ func makeExporterFactory(cfg *config.AgentConfig) func() (sdktrace.SpanExporter,
 	}
 }
 
+func createTLSConfig(cfg *config.AgentConfig) *tls.Config {
```
jcchavezs, line 116: If following the comment below we can also narrow down this to `*config.Reporting`.

instrumentation/opentelemetry/init.go
```diff
@@ -92,6 +96,14 @@ func makeExporterFactory(cfg *config.AgentConfig) func() (sdktrace.SpanExporter,
 			opts = append(opts, otlpgrpc.WithInsecure())
 		}
 
+		if len(cfg.GetReporting().GetCertFile().GetValue()) > 0 {
```
jcchavezs, line 99: `cfg.GetReporting().GetCertFile().GetValue()` is being called three times in this code, I believe we should move it into a variable.

instrumentation/opentelemetry/init_test.go
```diff
@@ -44,6 +44,19 @@ func ExampleRegisterService() {
 	defer shutdown()
 }
 
+func TestInitWithCertfileAndSecure(t *testing.T) {
+	cfg := config.Load()
+	cfg.ServiceName = config.String("my_example_svc")
```
jcchavezs, line 49: I might be missing something but I don't think we need all the config values being declared here and that sort of polutes the test. Can we just include those config that are required? otherwise it gives the impression this is required for the test.

jcchavezs (changes requested):
Looks great, just left some comments.

instrumentation/opentelemetry/init.go
tim-mwangi: fixed

instrumentation/opentelemetry/init.go
tim-mwangi: done

instrumentation/opentelemetry/init.go
tim-mwangi: done

instrumentation/opentelemetry/init_test.go
tim-mwangi: done

instrumentation/opentelemetry/init.go
jcchavezs: How about covering this function with tests?

instrumentation/opentelemetry/init.go
```diff
@@ -101,6 +114,35 @@ func makeExporterFactory(cfg *config.AgentConfig) func() (sdktrace.SpanExporter,
...
+		tlsConfig.RootCAs = createCaCertPoolFromFile(certFile)
+	}
+
+	return tlsConfig
+}
+
+// createCaCertPoolFromFile creates a CA Cert Pool from a file path containing
+// a raw CA certificate to verify a server certificate. The file path is the
+// reporting.cert_file config value.
+func createCaCertPoolFromFile(certFile string) *x509.CertPool {
```
jcchavezs, line 131: Shall we cover this function with tests?

jcchavezs (approved):
LGTM, would be great to cover that function I mentioned.

instrumentation/opentelemetry/init.go
tim-mwangi: It's already covered by the unit testing the caller :). Just the negative scenarios are not covered

tim-mwangi: done

PR #150, opened 2021-10-10
chore: upgrades OTel propagators to 1.0.0.

jcchavezs:
## Description

This PR upgrades OTel propagators aiming to solve the breaking change where `go.opentelemetry.io/contrib/propagators/b3` is either an own module and it isn't across versions.


tim-mwangi:
@jcchavezs tests are failing

PR #160, opened 2021-12-02
feat: adds support for body max processing size in bytes

jcchavezs:
Capture all bytes up to `max_processing_size` (1mb default) in memory and pass that through the blocking code.  Any data after `max_processing_size` should not be captured.  Only the value of `body_max_size_bytes` (128K default) should be added to the span attribute.

### DIFF
--- a/sdk/instrumentation/google.golang.org/grpc/server.go
+++ b/sdk/instrumentation/google.golang.org/grpc/server.go
@@ -96,9 +96,9 @@ func wrapHandler(
 			setTruncatedBodyAttribute("request", reqBody, int(dataCaptureConfig.BodyMaxSizeBytes.Value), span)
 
 			if md, ok := metadata.FromIncomingContext(ctx); ok {
-				var processingBody []byte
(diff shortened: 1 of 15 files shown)


sdk/instrumentation/google.golang.org/grpc/server.go
```diff
@@ -96,7 +96,11 @@ func wrapHandler(
 			setTruncatedBodyAttribute("request", reqBody, int(dataCaptureConfig.BodyMaxSizeBytes.Value), span)
 
 			if md, ok := metadata.FromIncomingContext(ctx); ok {
-				if filter.EvaluateBody(span, reqBody, md) {
+				var processingBody []byte
+				if int(dataCaptureConfig.BodyMaxProcessingSizeBytes.Value) > len(reqBody) {
```
davexroth, line 100: `reqBody` is the decoded value of the body right?  In which case, we should just use the same logic as https://github.com/hypertrace/goagent/pull/160/files#diff-c9de109efb7aa8920966082d6c5481c3324292b143637ff5c5f3f2174a5e1b9dR93

davexroth (approved):

PR #161, opened 2021-12-03
chore: adds initAsAdditional functionality to support coexistence with other tracers

jcchavezs:
This PR adds `InitAsAdditional` as a way to get goagent to coexist with other tracers in the OTel world.

See https://github.com/hypertrace/goagent/compare/adds_support_as_additional?expand=1#diff-4e82f63727576baabfc4f8a94f7944e1f70ae698d796b1e9736b35e92cfc38d0R22 for an example on the setup.


instrumentation/opentelemetry/init_additional.go
```diff
@@ -0,0 +1,136 @@
...
+
+// shieldAttrsSpan is a wrapper around a span that removes all attributes added by goagent as not
+// all the the tracing servers can handle the load of big attributes like body or headers.
+type shieldAttrsSpan struct {
+	trace.ReadOnlySpan
+}
+
+func (s *shieldAttrsSpan) Attributes() []attribute.KeyValue {
+	attrs := []attribute.KeyValue{}
+	for _, attr := range s.ReadOnlySpan.Attributes() {
```
davexroth, line 88: rather than having these in code, we should have some configuration to populate the list of attribute prefixes to filter out.

davexroth (approved):

PR #163, opened 2021-12-05
feat: adds support for enabled/disabled when it comes to wrappers.

jcchavezs:
WIP


davexroth:
> This PR improves performance of the interceptors/wrappers bypassing them when the agent is disabled. Notice propagation will still happen and noop tracing too. We will just not capture body or headers as they will end up in noop anyways.

This enabled flag needs to be the big red off switch a user sets to disable the go sdk if there are any issues, so we need to disable ourselves entirely when this is set.

I can see the use case for wanting to keep propagation with noop tracing, but that would be a different switch.

tim-mwangi:
Closing this one

PR #173, opened 2022-04-18
feat: provide API to read span attributes

ryanericson:


.github/workflows/pr-test.yml
```diff
@@ -40,11 +40,12 @@ jobs:
         uses: golangci/golangci-lint-action@v2
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
-          version: v1.29
-
-      - name: Check vanity import
-        run: |
-          make install-tools check-vanity-import
+          version: v1.42.1
+          # work around for "file exists" errors.
```
tim-mwangi, line 44: Why is this happening?

ryanericson: I have no idea tbh! It starts happening for hypertrace go agent CI. And then I discovered we put the same workaround for traceable go agent and TPA.

ryanericson: Even without my change the CI is failing with that file exists error.

tim-mwangi (approved):

PR #178, opened 2022-05-23
Add example outgoing call handler

prodion23:
## Description
Add an example outgoing http call

### DIFF
--- a/examples/http-server/main.go
+++ b/examples/http-server/main.go
@@ -32,6 +32,10 @@ func main() {
 		http.HandlerFunc(fooHandler),
 		"/foo",
 	))
+	r.Handle("/outgoing", hyperhttp.NewHandler(
+		http.HandlerFunc(outgoingCallHandler),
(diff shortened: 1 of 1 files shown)


examples/http-server/main.go
```diff
@@ -60,3 +64,31 @@ func fooHandler(w http.ResponseWriter, r *http.Request) {
...
 }
+
+func outgoingCallHandler(w, http.ResponseWriter, r *http.Request) {
+	client := http.Client{
+		Transport: hyperhttp.NewTransport(
+			http.DefaultTransport,
+		),
+	}
+
+	req, _ := http.NewRequestWithContext(r.Context(), "GET", "https://httpbin.org/get", nil)
```
tim-mwangi, line 75: Comment on this line and say that "In order to correlate the server call to this client call we need to pass the server request's context."

tim-mwangi:
huh. these tests may be unstable.

PR #181, opened 2022-06-24
Allow configurable content type capture

prodion23:
## Description
This adds the ability to configure captured content types. 

By default this value is "json", "x-www-form-urlencoded"


ryanericson:
@prodion23 can we go with agent-config route with a field like `allowedContentTypes`? By default it should just be `json` and `x-www-form-urlencoded`. Enabling xml data capture would mean specifying `json` , `x-www-form-urlencoded`, and maybe `xml` or up to the use case I guess if they want to restrict to just `application/xml`, or `application/soap+xml` or others.

sdk/instrumentation/net/http/contenttype.go
```diff
@@ -10,8 +11,14 @@ import (
 // - application/vnd.api+json
 // - application/x-www-form-urlencoded'
 var contentTypeAllowListLowerCase = []string{
-	"json",
-	"x-www-form-urlencoded",
+	"json", "x-www-form-urlencoded",
+}
+
+func SetContentTypeAllowList(config *agentconfig.AgentConfig) {
```
ryanericson, line 17: Do we need this call and `contentTypeAllowListLowerCase` global state? Can we just get it from the config? Seems like we have a `GetConfig()` API available.

tim-mwangi: +1 if it's easy to change from the hardcoded content types.

ryanericson (approved):

PR #183, opened 2022-08-04
 Allow setting of custom blocking rule

PrajwalKrishna:
## Description
Please include a summary of the change, motivation and context.

### Testing
Please describe the tests that you ran to verify your changes. Please summarize what did you test and what needs to be tested e.g. deployed and tested helm chart locally. 

### Checklist:

### Documentation


sdk/filter/filter.go
```diff
@@ -5,11 +5,11 @@ import "github.com/hypertrace/goagent/sdk"
 // Filter evaluates whether request should be blocked, `true` blocks the request and `false` continues it.
 type Filter interface {
 	// EvaluateURLAndHeaders can be used to evaluate both URL and headers
-	EvaluateURLAndHeaders(span sdk.Span, url string, headers map[string][]string) bool
+	EvaluateURLAndHeaders(span sdk.Span, url string, headers map[string][]string) (bool, int32)
```
ryanericson, line 8: Why not create a new structure like FilterResult?

sdk/filterutils/filter_result.go
```diff
@@ -0,0 +1,6 @@
+package filterutils
+
+type FilterResult struct {
```
ryanericson, line 3: Can we simply move FilterResult to filter.go?

examples/grpc-client/main.go
```diff
@@ -1,3 +1,4 @@
+//go:build ignore
```
PrajwalKrishna, line 1: I think these came up on running `make tidy`. Do we need this ?

sdk/filterutils/filter_result.go
PrajwalKrishna: This was leading to circular dependency as this would be needed by mocks as well.

ryanericson: Why would it as this type has no dependency to anything? Maybe I'm missing something.

examples/grpc-client/main.go
ryanericson: I don't think so. Can we revert all files that have nothing to do with the change? Not sure what the `go.sum` change is as well.

PrajwalKrishna: Okay so these are generated `make fmt`. Removing it for now.

sdk/filterutils/filter_result.go
PrajwalKrishna: Correct, but the module of the type needs to imported which will import other things.

PrajwalKrishna: `FilterResult` is needed by mock and we use `Mock` in `Multifilter_test` (which is part of filter package)  so if we put `FilterResult` in `filter.go` package we would have a circular dependency.

Either we move `FilterResult` to different package, or we have to move `Multifilter` to separate package.

sdk/filter/filter.go
```diff
@@ -1,15 +1,18 @@
 package filter // import "github.com/hypertrace/goagent/sdk/filter"
 
-import "github.com/hypertrace/goagent/sdk"
+import (
+	"github.com/hypertrace/goagent/sdk"
+	"github.com/hypertrace/goagent/sdk/filterutils"
```
tim-mwangi, line 5: Do we really need a `filterutils` package of can this just in the same filter package?

sdk/filterutils/filter_result.go
tim-mwangi, line 3: Can be in `sdk/filter` package as well. No need for a utils.

sdk/instrumentation/google.golang.org/grpc/server.go
```diff
@@ -104,8 +104,9 @@ func wrapHandler(
 				if int(dataCaptureConfig.BodyMaxProcessingSizeBytes.Value) < len(reqBody) {
 					processingBody = reqBody[:dataCaptureConfig.BodyMaxProcessingSizeBytes.Value]
 				}
-				if filter.EvaluateBody(span, processingBody, md) {
-					return nil, status.Error(codes.PermissionDenied, "Permission Denied")
+				filterResult := filter.EvaluateBody(span, processingBody, md)
+				if filterResult.Block {
+					return nil, status.Error(codes.Code(filterResult.ResponseStatusCode), "Permission Denied")
```
tim-mwangi, line 109: The status message will be different based on the status code. We probably need to map this correctly. We can have a static map for status code to status message based on https://developer.mozilla.org/en-US/docs/Web/HTTP/Status and just use that to get the correct message based on the status code.

sdk/filter/filter.go
PrajwalKrishna: https://github.com/hypertrace/goagent/pull/183#discussion_r939807930

sdk/instrumentation/google.golang.org/grpc/server.go
PrajwalKrishna: So what message should be I used incase the Status code is a non-standard one ?

tim-mwangi: Wait. You mean if the customer uses a status code that's is not one of the HTTP standard ones eg. 472? You can just have something generic like "Request Error". But for the well known ones, we can defined a global map of status code to status message and use that.

sdk/filter/filter.go
tim-mwangi: Aah I see. Ok. Let's use `"github.com/hypertrace/goagent/sdk/filter/result"` then instead of `"github.com/hypertrace/goagent/sdk/filterutils"`? I don't like utils since it ends up being a dumping ground of "other" functionality.

sdk/instrumentation/google.golang.org/grpc/status_text.go
```diff
@@ -0,0 +1,66 @@
...
+	case 414:
+		return "Request URI Too Long"
+	case 415:
+		return "Unsupported Media Type"
+	case 416:
+		return "Requested Range Not Satisfiable"
+	case 417:
+		return "Expectation Failed"
+	case 418:
+		return "I'm a teapot"
```
tim-mwangi, line 42: 😄

sdk/instrumentation/google.golang.org/grpc/status_text.go
```diff
@@ -0,0 +1,66 @@
+package grpc
+
+func StatusText(code int) string {
```
tim-mwangi, line 3: So the grpc package is the only one using this?

sdk/instrumentation/google.golang.org/grpc/status_text.go
```diff
@@ -0,0 +1,66 @@
...
+	switch code {
+	case 400:
+		return "Bad Request"
+	case 401:
+		return "Unauthorized"
+	case 402:
```
tim-mwangi, line 9: Nit: Also maybe we can turn this into a map variable and use the map instead of a switch. I think it's faster? If not found in map we can return "Request Error". To be honest I don't know what's faster between a map and a switch :)

tim-mwangi: Thanks for sharing these.

PR #184, opened 2022-08-25
feat: multipart/form-data data capture support

tim-mwangi:
## Description
Support capturing multipart/form-data request and response body. Note that we are base64 encoding these types of bodies since they may contain non utf8 bytes which would break the exporters.

### Testing
Added unit tests and verified locally using an example

### Checklist:
- [✅  ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/sdk/instrumentation/net/http/contenttype.go
+++ b/sdk/instrumentation/net/http/contenttype.go
@@ -5,11 +5,13 @@ import (
 	"strings"
 )
 
+const contentTypeHeaderKey string = "Content-Type"
+
 // ShouldRecordBodyOfContentType checks if the body is meant
 // to be recorded based on the content-type and the fact that body is
 // not streamed.
 func ShouldRecordBodyOfContentType(h HeaderAccessor) bool {
-	var contentTypeValues = h.Lookup("Content-Type") // "Content-Type" is the canonical key
+	var contentTypeValues = h.Lookup(contentTypeHeaderKey) // "Content-Type" is the canonical key
 	cfg := internalconfig.GetConfig().GetDataCapture()
 	if cfg == nil || cfg.GetAllowedContentTypes() == nil {
 		return false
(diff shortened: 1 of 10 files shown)


sdk/instrumentation/net/http/handler.go
```diff
@@ -87,16 +88,25 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			setTruncatedBodyAttribute("request", body, int(h.dataCaptureConfig.BodyMaxSizeBytes.Value), span,
+				isMultipartFormDataBody)
 		}
 
 		processingBody := body
 		if int(h.dataCaptureConfig.BodyMaxProcessingSizeBytes.Value) < len(body) {
 			processingBody = body[:h.dataCaptureConfig.BodyMaxProcessingSizeBytes.Value]
 		}
+		// if body is multipart/form-data, base64 encode it before passing it on to the filter
+		if isMultipartFormDataBody {
```
ryanericson, line 105: @tyroneyu does nginx also have processing size?

ryanericson (approved):

sdk/instrumentation/net/http/handler.go
tyroneyu: Yes, and the usage is the same as here.  It's the maximum size of the original body that we will take to process.  (not the base64 encoded string).

sdk/instrumentation/net/http/handler.go
```diff
@@ -87,16 +88,25 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 		defer r.Body.Close()
 
+		isMultipartFormDataBody := HasMultiPartFormDataContentTypeHeader(headerMapAccessor{r.Header})
+
 		// Only records the body if it is not empty and the content type
 		// header is not streamable
 		if len(body) > 0 {
-			setTruncatedBodyAttribute("request", body, int(h.dataCaptureConfig.BodyMaxSizeBytes.Value), span)
+			setTruncatedBodyAttribute("request", body, int(h.dataCaptureConfig.BodyMaxSizeBytes.Value), span,
```
tyroneyu, line 96: How does this work, the setTrucatedBodyAttribute has 4 arguments.

tyroneyu: Nevermind, that is confusing.  The logic looks correct.  There are 2 of those functions with the same name
`bodyattribute.setTruncatedBodyAttribute` and `body.setTruncatedBodyAttribute`.

minor: Perhaps it could all be combined into `body.setBodyAttributes()`.

tyroneyu (approved):

sdk/instrumentation/net/http/contenttype.go
```diff
@@ -43,3 +45,15 @@ func ShouldRecordBodyOfContentType(h HeaderAccessor) bool {
 	}
 	return false
 }
+
+// HasMultiPartFormDataContentTypeHeader returns true if the Content-Type header is
+// multipart/form-data. false otherwise.
+func HasMultiPartFormDataContentTypeHeader(h HeaderAccessor) bool {
+	var contentTypeValues = h.Lookup(contentTypeHeaderKey)
+	for _, contentTypeValue := range contentTypeValues {
```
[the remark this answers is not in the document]
tim-mwangi: i have to check.

sdk/instrumentation/net/http/handler.go
tim-mwangi: they should all be in different packages.

PR #189, opened 2022-10-03
Update golangci lint - add verbose logging

prodion23:
## Description
Add verbose lint logging


prodion23:
@tim-mwangi still unsure how this fails on main but not on prs.  It's the same lint command running.. 🤔

PR #191, opened 2022-11-01
Add round robin load balancing otlp config, set resolve as dns

prodion23:
This adds an opt-in feature to set grpc round robin load balancing & changing the default resolver to DNS instead of passthrough.


tim-mwangi (approved):

instrumentation/opentelemetry/init.go
```diff
@@ -111,6 +112,11 @@ func makeExporterFactory(cfg *config.AgentConfig) func() (sdktrace.SpanExporter,
 			}
 		}
 
+		if cfg.Reporting.GetEnableGrpcLoadbalancing().GetValue() {
```
ryanericson, line 115: @prodion23 do we not need to add default value in https://github.com/hypertrace/goagent/blob/4823a3e6efd3dc1fc622331ec6b8a7c7e27e9716/config/defaults.go#L34?

tim-mwangi: Very good catch.

prodion23: Thanks for catching that, will fix

PR #195, opened 2022-11-30
feat: create a startup span in init

tim-mwangi:
## Description
Important to indicate when an instrumented app starts up

### Testing
Tested locally using an instrumented app.

### Checklist:

### DIFF
--- a/instrumentation/opentelemetry/init.go
+++ b/instrumentation/opentelemetry/init.go
@@ -226,7 +226,7 @@ func InitWithSpanProcessorWrapper(cfg *config.AgentConfig, wrapper SpanProcessor
 
 	// Startup span
 	_, span, ender := startSpanFn(context.Background(), "startup", &sdk.SpanOptions{})
-	span.SetAttribute("hypertrace.goagent.startup", true)
+	span.SetAttribute("hypertrace.agent.startup", true)
 	ender()
 
 	return func() {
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry/init.go
```diff
@@ -220,6 +220,15 @@ func InitWithSpanProcessorWrapper(cfg *config.AgentConfig, wrapper SpanProcessor
 	globalSampler = sampler
 	initialized = true
 
+	startSpanFn := startSpan(func() trace.TracerProvider {
+		return tp
+	})
+
+	// Startup span
+	_, span, ender := startSpanFn(context.Background(), "startup", &sdk.SpanOptions{})
+	span.SetAttribute("hypertrace.goagent.startup", true)
```
prodion23, line 229: Hmm, do you think this should be lang specific?  

`hypertrace.startup` or similar? 

Maybe not even valid concern but just to float it.
if a platform will have to look for these specific attributes something more generic could be better, so that we don't to update or maintain a separate list of `hypertrace.some-lang.startup` attrs to look for)

If we were just to filter on `hypertrace.startup` we could still tell languages apart by the `telemetry.sdk.language` attr.

Thoughts?

tim-mwangi: Good call. I will rename to `hypertrace.agent.startup`

prodion23 (approved):

ryanericson (approved):

PR #199, opened 2023-01-05
Add "service.instance.id" to every span

prodion23:
## Description

This adds a `service.instance.id` to every span.  

I tried adding this as a processor; however that didn't seem to work with our `InitWithSpanProcessorWrapper` I also tried adding it as a starting attribute using `trace.WithAttributes` however; those are filtered out during sampling.

Instead we just add it to each span during startSpan. If we'd prefer a different approach happy to revisit.(this seems like the most minimally invasive approach i think)

From docs about this attribute: 
> [3]: MUST be unique for each instance of the same service.namespace,service.name pair (in other words service.namespace,service.name,service.instance.id triplet MUST be globally unique). The ID helps to distinguish instances of the same service that exist at the same time (e.g. instances of a horizontally scaled service). It is preferable for the ID to be persistent and stay the same for the lifetime of the service instance, however it is acceptable that the ID is ephemeral and changes during important lifetime events for the service (e.g. service restarts). If the service has no inherent unique ID that can be used as the value of this attribute it is recommended to generate a random Version 1 or Version 4 RFC 4122 UUID (services aiming for reproducible UUIDs may also use Version 5, see RFC 4122 for more recommendations).

Specifically of note - I think we are okay with this being ephemeral for our specific use case.  (A restarting service will have different id)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,8 +13,8 @@ require (
 	github.com/mattn/go-sqlite3 v1.14.4
 	github.com/ngrok/sqlmw v0.0.0-20200129213757-d5c93a81bec6
 	github.com/openzipkin/zipkin-go v0.4.0
+	github.com/satori/go.uuid v1.2.1-0.20181028125025-b2ce2384e17b
 	github.com/stretchr/testify v1.8.1
-	github.com/satori/go.uuid v1.2.0
 	go.opencensus.io v0.23.0
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.35.0
 	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.35.0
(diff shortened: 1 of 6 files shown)


ryanericson (approved):

ryanericson (approved):

tim-mwangi:
@prodion23 I see that we set this as a span attribute. Any reason why we didn't set it as a resource attribute?

ryanericson:
@tim-mwangi because resource attributes aren't persisted and queryable yet. https://traceableai.atlassian.net/browse/ENG-24793?focusedCommentId=47515, so making it as span attribute will make it queryable in explorer/pinot.

tim-mwangi:
> @tim-mwangi because resource attributes aren't persisted and queryable yet. https://traceableai.atlassian.net/browse/ENG-24793?focusedCommentId=47515, so making it as span attribute will make it queryable in explorer/pinot.

yeah @prodion23 explained on slack.

PR #200, opened 2023-01-24
feat: otel go metrics setup

tim-mwangi:
## Description
Set up otel go metrics reporting and add a RequestCount metric for the http handler instrumentation. This should enable metrics to be created and exported just like spans are. By default we will export to the otlp endpoint spans are exported to but this can be overridden using a metrics endpoint config.

### Testing
Tested locally.

### Checklist:


tim-mwangi:
Still WIP

sdk/instrumentation/net/http/handler.go
```diff
@@ -12,14 +12,31 @@ import (
 	"github.com/hypertrace/goagent/sdk/filter"
 	internalconfig "github.com/hypertrace/goagent/sdk/internal/config"
 	"github.com/hypertrace/goagent/sdk/internal/container"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
```
tim-mwangi, line 15: I need to move this otel specific logic out of the sdk since this is meant to be a hypertrace sdk that can be adapted by any telemetry framework. But can do that separately.

prodion23 (approved):

instrumentation/opentelemetry/init.go
```diff
@@ -321,6 +382,42 @@ func RegisterServiceWithSpanProcessorWrapper(serviceName string, resourceAttribu
...
 
+func initializeMetrics(cfg *config.AgentConfig) func() {
+	if cfg.GetTelemetry() == nil || !cfg.GetTelemetry().GetMetricsEnabled().GetValue() {
+		return func() {}
+	}
+
+	metricsExporterFactory := makeMetricsExporterFactory(cfg)
+	metricsExporter, err := metricsExporterFactory()
+	if err != nil {
+		log.Fatal(err)
```
ryanericson, line 393: This means if TPA is down instrumented app won't be able to start? Better to fail silently?

tim-mwangi: If the metrics/traces collector is down and you bring up the go app with this agent in it, it comes up without issue. You do get context deadline messages since you cannot export anything. After you bring up the collector the spans and metrics start being sent.

ryanericson (approved):

PR #201, opened 2023-01-27
feat: use a modified batch span processor that creates metrics

tim-mwangi:
## Description
This is to track spans received and dropped by the processor. Currently the processor silently drops spans when the queue is full which is quite difficult to debug. There's a debug log but then this means the user modifying their code to enable it and the logs being quite noisy. Using metrics is much better.

### Testing
Tested locally.

### Checklist:
- [ ✅ ] I have added tests that prove my fix is effective or that my feature works


instrumentation/opentelemetry/batchspanprocessor/batch_span_processor.modified.go
```diff
@@ -0,0 +1,434 @@
...
+		ExportTimeout:      time.Duration(BatchSpanProcessorExportTimeout(DefaultExportTimeout)) * time.Millisecond,
+		MaxQueueSize:       maxQueueSize,
+		MaxExportBatchSize: maxExportBatchSize,
+	}
+	for _, opt := range options {
+		opt(&o)
+	}
+
+	// Setup metrics
+	mp := metricglobal.MeterProvider()
```
ryanericson, line 91: what happens if metrics is disabled? Will it no op?

tim-mwangi: Well, it's able to handle it without explicitly using the NoopMetricProvider. Either way it returns a meter. See https://github.com/open-telemetry/opentelemetry-go/blob/ec13377b6b7cea9ba5e6a2233ad934a411459a58/metric/internal/global/meter.go#L69 . Which is why I wonder if we should use our modified bsp throughout. Right now if metrics is disabled, I switch to the original one but I don't think we need to do that.

ryanericson (approved):

PR #204, opened 2023-02-01
Bump to otel 1.12.0

tyroneyu:
## Description

Bump otel to 1.12.0


go.mod
```diff
@@ -28,25 +28,24 @@ require (
 	go.opentelemetry.io/otel/metric v0.31.0
 	go.opentelemetry.io/otel/sdk v1.10.0
 	go.opentelemetry.io/otel/trace v1.10.0
-	google.golang.org/grpc v1.49.0
```
tim-mwangi, line 31: Need to bump metrics as well to the corresponding version: i.e go.opentelemetry.io/otel/metric I believe it's v0.35.0 https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.12.0

tyroneyu: Thanks! that was the hint that I needed.  I also had to bump up the `contrib` versions as well.

instrumentation/opentelemetry/init.go
```diff
@@ -399,31 +393,25 @@ func initializeMetrics(cfg *config.AgentConfig) func() {
...
+	meterProvider := metric.NewMeterProvider(metric.WithReader(periodicReader), metric.WithResource(metricResources))
+	otelmetricglobal.SetMeterProvider(meterProvider)
 
 	return func() {
-		err := metricsPusher.Stop(context.Background())
+		err := periodicReader.Shutdown(context.Background())
+		if err != nil {
+			log.Printf("an error while calling metrics reader shutdown: %v", err)
+		}
+		err = meterProvider.Shutdown(context.Background())
```
tim-mwangi, line 408: Should we invert the order of these 2 shutdowns since periodic reader was created first?

.github/workflows/pr-test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        go: ["1.17", "1.18", "1.19"]
+        go: ["1.18", "1.19"]
```
tim-mwangi, line 16: Add "1.20"

instrumentation/opentelemetry/init.go
tyroneyu: I did it this way so that we would read all the latest metrics and push them out.  If it is flipped we will not have the last few right?

tim-mwangi: Not sure I follow. If you shutdown the meter provider first, it means no one can create new meters dependent on the periodic reader we created.

tyroneyu: I'll switch it. I got thrown off in one of their tests. I read "reader" as just the collect.  It is both the collector and exporter.

Done.

.github/workflows/pr-test.yml
tyroneyu: Done

For future reference, where did you read the 1.20 support?  I used this link for reference to know 1.17 was gone. I did not see 1.20.
https://github.com/open-telemetry/opentelemetry-go

tim-mwangi: In the compatibility it says latest version - 2. https://github.com/open-telemetry/opentelemetry-go#compatibility

tim-mwangi (approved):

.github/workflows/pr-test.yml
tyroneyu: For posterity, otel hasn't updated their compatibiity chart nor their CI yet. It still only runs golang 1.18/1.19.  However, golang 1.20 was released Feb 1st and otel points to https://go.dev/doc/devel/release#policy to say they support the last 2 golang versions. 

It does not hurt to add this into our CI right now. 

Done.

PR #212, opened 2023-03-03
chore: make it easier to pass resource agent name and version

tim-mwangi:
## Description
This will make it easier for consumers of this module to pass their own agent names and versions.

### Testing
Unit tests and tested locally

### Checklist:

### DIFF
--- a/instrumentation/opentelemetry/init.go
+++ b/instrumentation/opentelemetry/init.go
@@ -208,12 +208,13 @@ func createCaCertPoolFromFile(certFile string) *x509.CertPool {
 // Init initializes opentelemetry tracing and returns a shutdown function to flush data immediately
 // on a termination signal.
 func Init(cfg *config.AgentConfig) func() {
-	return InitWithSpanProcessorWrapper(cfg, nil)
+	return InitWithSpanProcessorWrapper(cfg, nil, versionInfoResourceAttributes)
 }
 
 // InitWithSpanProcessorWrapper initializes opentelemetry tracing with a wrapper over span processor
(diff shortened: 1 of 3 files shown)


instrumentation/opentelemetry/init.go
```diff
@@ -208,12 +208,13 @@ func createCaCertPoolFromFile(certFile string) *x509.CertPool {
...
 // on a termination signal.
 func Init(cfg *config.AgentConfig) func() {
-	return InitWithSpanProcessorWrapper(cfg, nil)
+	return InitWithSpanProcessorWrapper(cfg, nil, versionInfoResourceAttributes)
 }
 
 // InitWithSpanProcessorWrapper initializes opentelemetry tracing with a wrapper over span processor
 // and returns a shutdown function to flush data immediately on a termination signal.
-func InitWithSpanProcessorWrapper(cfg *config.AgentConfig, wrapper SpanProcessorWrapper) func() {
+func InitWithSpanProcessorWrapper(cfg *config.AgentConfig, wrapper SpanProcessorWrapper,
```
tyroneyu, line 216: Just curious, what is the advantage of passing around the func rather than the slice? 

Secondary is, can the versionInfo be const?

tim-mwangi: I could make it a var which will be an attribute array.

tim-mwangi: Done

prodion23 (approved):

tyroneyu (approved):

PR #214, opened 2023-06-28
Update dependencies

prodion23:
Drops 1.18 support


examples/grpc-client/main.go
```diff
@@ -1,3 +1,4 @@
+//go:build ignore
```
tim-mwangi, line 1: oh these are now needed?

prodion23: Hmm,I think these were added with fmt, but we probably don't need both lines 1 & 2

tim-mwangi: cool. let's remove them.

tim-mwangi (approved):

tim-mwangi:
Hey @prodion23 can we file JIRA to look at some of these gosec issues? https://github.com/hypertrace/goagent/pull/214/checks?check_run_id=14688939422 Thanks.

PR #217, opened 2023-11-05
 adding generic attributes interface 

varkey98:
## Description
Adding a generic GetAll method for attributes so that all the attributes can be queried at once

### Checklist:

### Documentation

### DIFF
--- a/instrumentation/opentelemetry/span.go
+++ b/instrumentation/opentelemetry/span.go
@@ -32,9 +32,10 @@ func (l *AttributeList) GetValue(key string) interface{} {
 }
 
 func (l *AttributeList) GetAll() []sdk.Attribute {
-	attributes := make([]sdk.Attribute, len(l.attrs))
-	for _, attr := range l.attrs {
-		attributes = append(attributes, sdk.Attribute{Key: string(attr.Key), Value: attr.Value.AsInterface()})
+	size := len(l.attrs)
+	attributes := make([]sdk.Attribute, size)
+	for i := 0; i < size; i++ {
+		attributes[i] = sdk.Attribute{Key: string(l.attrs[i].Key), Value: l.attrs[i].Value.AsInterface()}
 	}
 	return attributes
 }
--- a/instrumentation/opentelemetry/span.go
+++ b/instrumentation/opentelemetry/span.go
@@ -31,6 +31,14 @@ func (l *AttributeList) GetValue(key string) interface{} {
 	return nil
 }
 
+func (l *AttributeList) GetAll() []sdk.Attribute {
+	attributes := make([]sdk.Attribute, len(l.attrs))
+	for _, attr := range l.attrs {
+		attributes = append(attributes, sdk.Attribute{Key: string(attr.Key), Value: attr.Value.AsInterface()})
+	}
+	return attributes
+}
+
 var _ sdk.Span = (*Span)(nil)
 
 type Span struct {
--- a/instrumentation/opentelemetry/span_test.go
+++ b/instrumentation/opentelemetry/span_test.go
@@ -109,3 +109,23 @@ func TestGetAttributes(t *testing.T) {
 	assert.Equal(t, "string_value", attrs.GetValue("string_key"))
 	assert.Equal(t, nil, attrs.GetValue("non_existent"))
 }
+
+func TestGetAllAttributes(t *testing.T) {
+	sampler := sdktrace.AlwaysSample()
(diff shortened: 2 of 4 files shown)


varkey98:
We would need to use fmt.Sprintf before passing the values to the filter, shall we just store the attributes as strings itself in that case? @ryanericson @tim-mwangi

instrumentation/opentelemetry/span.go
```diff
@@ -31,6 +31,14 @@ func (l *AttributeList) GetValue(key string) interface{} {
 	return nil
 }
 
+func (l *AttributeList) GetAll() []sdk.Attribute {
+	attributes := make([]sdk.Attribute, len(l.attrs))
+	for _, attr := range l.attrs {
+		attributes = append(attributes, sdk.Attribute{Key: string(attr.Key), Value: attr.Value.AsInterface()})
```
tim-mwangi, line 37: Oh this is tricky in golang. I think it will produce an array with twice the intended length and with first of the array full of empty values. append does not start at index 0.

varkey98: Thanks, I thought the size we specify was just for memory allocation, didnt realise we get a filled slice. Updated it

sdk/internal/mock/span.go
```diff
@@ -29,6 +29,14 @@ func (l *AttributeList) GetValue(key string) interface{} {
 	return l.attrs[key]
 }
 
+func (l *AttributeList) GetAll() []sdk.Attribute {
+	attributes := make([]sdk.Attribute, len(l.attrs))
```
tim-mwangi, line 33: Also make the same change here :)

instrumentation/opentelemetry/span_test.go
```diff
@@ -109,3 +109,24 @@ func TestGetAttributes(t *testing.T) {
...
+	tp := sdktrace.NewTracerProvider(
+		sdktrace.WithSampler(sampler),
+	)
+	otel.SetTracerProvider(tp)
+	_, s, _ := StartSpan(context.Background(), "test_span", &sdk.SpanOptions{})
+	s.SetAttribute("k1", "v1")
+	s.SetAttribute("k2", 200)
+	attrs := s.GetAttributes().GetAll()
+
+	assert.Equal(t, 3, len(attrs))
```
tim-mwangi, line 124: Why 3?

varkey98: `service.instance.id` is already there as an attribute

sdk/internal/mock/span.go
varkey98: sorry, missed that

instrumentation/opentelemetry/span_test.go
tim-mwangi:
```suggestion
        // service.instance.id is added implicitly in StartSpan so 3 attributes will be present.
	assert.Equal(t, 3, len(attrs))
```

sdk/internal/mock/span.go
```diff
@@ -29,6 +29,16 @@ func (l *AttributeList) GetValue(key string) interface{} {
 	return l.attrs[key]
 }
 
+func (l *AttributeList) GetAll() []sdk.Attribute {
+
+	attributes := make([]sdk.Attribute, len(l.attrs))
+	i := 0
+	for key, value := range l.attrs {
+		attributes[i] = sdk.Attribute{Key: key, Value: value}
```
tim-mwangi, line 37:
```suggestion
		attributes[i] = sdk.Attribute{Key: key, Value: value}
		i++
```

tim-mwangi (approved):
lgtm. @ryanericson please take a look as well since you were the original author for this :)

varkey98:
One of you have to merge it as well 
image

instrumentation/opentelemetry/span.go
```diff
@@ -31,6 +31,15 @@ func (l *AttributeList) GetValue(key string) interface{} {
 	return nil
 }
 
+func (l *AttributeList) GetAll() []sdk.Attribute {
```
ryanericson, line 34: @varkey98 If this is for the purpose of being able to access all attributes in filter impl, can we find a way to just expose a way to iterate over each attribute in the list? Rather than incurring overhead of another slice and copying over to that slice. All attributes are already contained in the list.

varkey98: Got it, let me try. I created a new type because of the generic nature of `AttributeList`

PR #219, opened 2023-11-10
ENG-37132: Adding bodyAttribute functions to support already truncated bodies

puneet-traceable:
## Description
This PR adds functions to set bodies in span in case bodies are already truncated.
Currently for body attributes `SetTruncatedBodyAttribute` and `SetTruncatedEncodedBodyAttribute` functions are supported, but to use those users will need to read complete body and then let these functions do the truncation.
to support reading limited sized bodies, new functions are added which truncated attribute.

### Testing
Will be updating unit tests.

### Checklist:
- [ ] My changes generate no new warnings

### DIFF
--- a/sdk/instrumentation/bodyattribute/bodyattribute.go
+++ b/sdk/instrumentation/bodyattribute/bodyattribute.go
@@ -17,13 +17,11 @@ func SetTruncatedBodyAttribute(attrName string, body []byte, bodyMaxSize int, sp
 	}
 
 	if bodyLen <= bodyMaxSize {
-		span.SetAttribute(attrName, string(body))
+		SetBodyAttribute(attrName, body, false, span)
 		return
 	}
 
-	truncatedBody := body[:bodyMaxSize]
-	span.SetAttribute(fmt.Sprintf("%s.truncated", attrName), true)
-	span.SetAttribute(attrName, string(truncatedBody))
+	SetBodyAttribute(attrName, body[:bodyMaxSize], true, span)
 }
 
 // SetTruncatedEncodedBodyAttribute is like SetTruncatedBodyAttribute above but also base64 encodes the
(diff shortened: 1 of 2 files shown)


puneet-traceable:
@ryanericson @tim-mwangi not sure on our take of making breaking changes in this repo. I have added new functions for now, but if we are ok making breaking changes I can change existing functions. Also, will add tests based on inputs on these lines.

sdk/instrumentation/bodyattribute/bodyattribute.go
```diff
@@ -44,3 +44,48 @@ func SetTruncatedEncodedBodyAttribute(attrName string, body []byte, bodyMaxSize
 	span.SetAttribute(fmt.Sprintf("%s.truncated", attrName), true)
 	span.SetAttribute(attrName+".base64", base64.RawStdEncoding.EncodeToString(truncatedBody))
 }
+
+// SetPossiblyTruncatedBodyAttribute sets the body as a span attribute.
+// When body is being truncated, or already truncated we also add a second attribute suffixed by `.truncated` to
+// make it clear to the user, body has been modified.
+func SetPossiblyTruncatedBodyAttribute(attrName string, body []byte, bodyMaxSize int, truncated bool, span sdk.Span) {
```
ryanericson, line 51: Does this one need to have truncation logic? It can just be used when client code wants to handle truncation on its own in which case no truncation logic is needed so function can just be `setBodyAtttribute(attrName string, body []byte, truncated bool, span sdk.Span)`.

ryanericson:
> @ryanericson @tim-mwangi not sure on our take of making breaking changes in this repo. I have added new functions for now, but if we are ok making breaking changes I can change existing functions. Also, will add tests based on inputs on these lines.

It's ok to break compat but not sure if we need to go that way and function can be confusing IMO. Perhaps best to have 2 versions where one can do truncation inside, and another never does truncation but can put `.truncated` attribute i.e. it assumes truncation is handled outside.

sdk/instrumentation/bodyattribute/bodyattribute.go
puneet-traceable: updated and used the `SetBodyAtttribute` from `SetTruncatedBodyAttribute`

puneet-traceable:
> > @ryanericson @tim-mwangi not sure on our take of making breaking changes in this repo. I have added new functions for now, but if we are ok making breaking changes I can change existing functions. Also, will add tests based on inputs on these lines.
> 
> It's ok to break compat but not sure if we need to go that way and function can be confusing IMO. Perhaps best to have 2 versions where one can do truncation inside, and another never does truncation but can put `.truncated` attribute i.e. it assumes truncation is handled outside.

yeah keeping the two but trying to reuse one by the other.

varkey98 (approved):

ryanericson (approved):

sdk/instrumentation/bodyattribute/bodyattribute.go
```diff
@@ -17,30 +17,55 @@ func SetTruncatedBodyAttribute(attrName string, body []byte, bodyMaxSize int, sp
...
+	SetBodyAttribute(attrName, body[:bodyMaxSize], true, span)
 }
 
 // SetTruncatedEncodedBodyAttribute is like SetTruncatedBodyAttribute above but also base64 encodes the
 // body. This is usually due to non utf8 bytes in the body eg. for multipart/form-data content type.
 // The body attribute name has a ".base64" suffix.
 func SetTruncatedEncodedBodyAttribute(attrName string, body []byte, bodyMaxSize int, span sdk.Span) {
 	bodyLen := len(body)
-	if bodyLen == 0 {
+	if len(body) == 0 {
```
tim-mwangi, line 32: why not use the variable `bodyLen`?

puneet-traceable: will fix it, missed it in a revert.

PR #220, opened 2023-11-11
adding iterator interface and removing GetAll() method from attribute list

varkey98:
Adding an iterator interface on AttributeList so that the iterator can be queried without creating an additional slice. 
Continuation of this pr: https://github.com/hypertrace/goagent/pull/217
ref: https://github.com/hypertrace/goagent/pull/217/files#r1385540220

### Checklist:

### DIFF
--- a/instrumentation/opentelemetry/span.go
+++ b/instrumentation/opentelemetry/span.go
@@ -60,6 +60,14 @@ func (l *AttributeList) GetIterator() sdk.Iterator {
 	}
 }
 
+func (l *AttributeList) IterateItems(yield func(attr sdk.Attribute) bool) {
+	for _, attr := range l.attrs {
+		if !yield(sdk.Attribute{Key: string(attr.Key), Value: attr.Value.AsInterface()}) {
+			return
+		}
+	}
+}
+
 var _ sdk.Span = (*Span)(nil)
 
 type Span struct {
--- a/instrumentation/opentelemetry/span.go
+++ b/instrumentation/opentelemetry/span.go
@@ -15,6 +15,27 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
+var _ sdk.Iterator = (*Iterator)(nil)
+
+type Iterator struct {
+	size  int
+	curr  int
+	attrs []attribute.KeyValue
+}
+
+func (i *Iterator) HasNext() bool {
+	return i.curr < i.size
+}
+
+func (i *Iterator) Next() sdk.Attribute {
+	attr := sdk.Attribute{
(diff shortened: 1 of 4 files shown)


sdk/span.go
```diff
@@ -10,9 +10,14 @@ type Attribute struct {
 	Value interface{}
 }
 
+type Iterator interface {
+	HasNext() bool
+	Next() Attribute
```
varkey98, line 15: @ryanericson @tim-mwangi Still returning the generic interface Attribute since we want to fetch the key as well. Or we can return the otel key-value pair directly

varkey98: Or we can keep on using the same `Attribute` object so that it's initialised only once

ryanericson: Why not do something like: https://bitfieldconsulting.com/golang/iterators#:~:text=A%20Golang%20iterator%20is%20a,value%20returned%20by%20the%20iterator.

varkey98: But here also what would we pass as the argument of the yield function? If we want to keep the generality of the attribute, we would have to return a common struct right?
like 
```
func (l *AttributeList) IterateItems(yield func(attr Attribute) bool) {
	for _, attr := range l.attrs {
		if !yield(attr) {
			return
		}
	}
}
```

varkey98: Thanks, updated the pr. Didnt know we could make an iterator like this

instrumentation/opentelemetry/span.go
```diff
@@ -31,13 +31,12 @@ func (l *AttributeList) GetValue(key string) interface{} {
 	return nil
 }
 
-func (l *AttributeList) GetAll() []sdk.Attribute {
-	size := len(l.attrs)
-	attributes := make([]sdk.Attribute, size)
-	for i := 0; i < size; i++ {
-		attributes[i] = sdk.Attribute{Key: string(l.attrs[i].Key), Value: l.attrs[i].Value.AsInterface()}
+func (l *AttributeList) Iterate(yield func(key string, value interface{}) bool) {
```
tim-mwangi, line 34:
```suggestion
// Iterate loops through the attributes list and applies the yield function on each attribute. If the
// yield function returns false, we exit the loop.
func (l *AttributeList) Iterate(yield func(key string, value interface{}) bool) {
```

varkey98: adding it in the interface

tim-mwangi (approved):

ryanericson (approved):

varkey98:
@ryanericson @tim-mwangi Please merge this pr, I cant merge since I dont have write access

PR #223, opened 2023-11-16
feat: update filter interface to process only based on spans

varkey98:
## Description
Updating the filter interface to have only a single method, `Evaluate`. And when calling evaluate, all the relevant attributes/headers are expected to be present as part of the span

### Checklist:

### DIFF
--- a/sdk/instrumentation/google.golang.org/grpc/server.go
+++ b/sdk/instrumentation/google.golang.org/grpc/server.go
@@ -95,14 +95,6 @@ func wrapHandler(
 
 		if dataCaptureConfig.RpcMetadata.Request.Value {
 			setAttributesFromRequestIncomingMetadata(ctx, span)
-
-			// TODO: decide what should be passed as URL in GRPC
-			if !dataCaptureConfig.RpcBody.Request.Value {
-				filterResult := filter.Evaluate(span)
-				if filterResult.Block {
-					return nil, status.Error(StatusCode(int(filterResult.ResponseStatusCode)), StatusText(int(filterResult.ResponseStatusCode)))
-				}
-			}
 		}
 
 		reqBody, err := marshalMessageableJSON(req)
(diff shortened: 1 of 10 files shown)


sdk/instrumentation/google.golang.org/grpc/server.go
```diff
@@ -94,32 +93,26 @@ func wrapHandler(
 
 		setSchemeAttributes(ctx, span)
 
-		reqBody, err := marshalMessageableJSON(req)
-		if dataCaptureConfig.RpcBody.Request.Value &&
-			len(reqBody) > 0 && err == nil {
-			setTruncatedBodyAttribute("request", reqBody, int(dataCaptureConfig.BodyMaxSizeBytes.Value), span)
+		if dataCaptureConfig.RpcMetadata.Request.Value {
```
varkey98, line 96: @ryanericson @puneet-traceable @tim-mwangi This reordering was required because we now only have a single method in filter interface and the scenario of blocking based on headers was failing since the only call to filter was done on body first.

I think we should make similar changes to all instrumentations so that our issue of allow rule not taking precedence over blocking rules also will be solved. 
Or we can pass flowvariable to Evaluate method to specify if we're calling for header eval or body eval

tim-mwangi: So libtraceable interface only has one method called `Evaluate` now? Do the conditions that had it split up into various methods no longer apply?

varkey98: Yes, libtraceable interface has only one method `Evaluate` now. 

> Do the conditions that had it split up into various methods no longer apply?
This I'm not really sure. I looked through the code and couldn't find any reason why we had separate calls (maybe for consistency across agents). In both the cases we call the delegate after the body capture. I'll test with an actual app over the weekend.

varkey98: @tim-mwangi I tested with both a `net/http` app and and a `grpc` app, both of them are blocking with and w/o body

ryanericson (approved):

sdk/instrumentation/net/http/handler.go
```diff
@@ -103,25 +93,14 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 				isMultipartFormDataBody)
 		}
 
-		processingBody := body
-		if int(h.dataCaptureConfig.BodyMaxProcessingSizeBytes.Value) < len(body) {
-			processingBody = body[:h.dataCaptureConfig.BodyMaxProcessingSizeBytes.Value]
```
varkey98, line 108: `dataCaptureConfig.BodyMaxProcessingSizeBytes` config is also becoming obsolete as we are using the body stored as span attribute for filter evaluation

ryanericson: Yes, I'm ok with that.

ryanericson: Can you help open a ticket to deprecate body max processing size throughout? Thanks!

varkey98: yes, :https://traceableai.atlassian.net/browse/ENG-37604

PR #224, opened 2023-11-23
feat: adding length method to attribute list and updating http.url attribute in net/http instrumentation

varkey98:
## Description
1. adding a length method to attribute list
2. Updating the http.url attribute in net/http instrumentation in case if its not the complete url

### Checklist:

### DIFF
--- a/sdk/instrumentation/net/http/handler.go
+++ b/sdk/instrumentation/net/http/handler.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"io"
 	"net/http"
+	"strings"
 
 	config "github.com/hypertrace/agent-config/gen/go/v1"
 	"github.com/hypertrace/goagent/sdk"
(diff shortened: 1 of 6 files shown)


sdk/instrumentation/net/http/handler.go
```diff
@@ -65,7 +66,11 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	url := r.URL.String()
-	span.SetAttribute("http.url", url)
+	if strings.Contains(url, "://") {
```
varkey98, line 69: fixing this issue in the instrumentation itself: https://github.com/Traceableai/goagent-src/blob/2320366e224b742d023236680e7b460e4a17630f/filter/traceable/filter.go#L352

tim-mwangi (approved):

PR #229, opened 2024-05-23
feat: config for custom batch span processor

tim-mwangi:
## Description
We need to be able to pick whether we use out customer bsp https://github.com/hypertrace/goagent/tree/main/instrumentation/opentelemetry/batchspanprocessor or use the stock one in the opentelemetry go repo.

### Testing
Unit and local tests

### Checklist:

### DIFF
--- a/instrumentation/opentelemetry/init.go
+++ b/instrumentation/opentelemetry/init.go
@@ -34,6 +34,7 @@ import (
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
 	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/noop"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/resolver"
 )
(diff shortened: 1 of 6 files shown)


tim-mwangi:
Oh I missed some places eg. `RegisterServiceWithSpanProcessorWrapper()`. Will fix.

ryanericson (approved):

instrumentation/opentelemetry/init.go
```diff
@@ -243,14 +251,15 @@ func InitWithSpanProcessorWrapper(cfg *config.AgentConfig, wrapper SpanProcessor
...
 	exporterFactory = makeExporterFactory(cfg)
+	configFactory = makeConfigFactory(cfg)
 
 	exporter, err := exporterFactory()
 	if err != nil {
 		log.Fatal(err)
 	}
 
 	sp := modbsp.CreateBatchSpanProcessor(
-		cfg.GetTelemetry() != nil && cfg.GetTelemetry().GetMetricsEnabled().GetValue(), // metrics enabled
```
puneet-traceable, line 253: Just to understand this more, why was this conditional to run BatchSpanProcesor until now? What I am trying to understand is what would happen if customBsp is false and metrics are enabled? Will non-modified batchSpanProcessor work?

tim-mwangi: The only reason for the custom BatchSpanProcessor(BSP) is to be able to collect span metrics - received, dropped and unsampled - as they hit the exporter. So if metrics is disabled, there's no need for it.
As if this PR if custom BSP is false and metrics are enabled, we use the stock BSP. I think this behavior will remain even in the recover from panic PR. And I'm thinking I will remove the metrics enabled requirement in that PR so that we can recover even if metrics are disabled and panic happens when exporting the traces.

PR #236, opened 2024-07-24
feat: updating filter result to add decorations

varkey98:
## Description
Adding a Decoration field to FilterResult. The Decoration struct can have mutations defined on request or response objects which should be applied on a filter result match. Starting with request header injections, in which a list of headers can be specified to be injected on a filter evaluation match 

### Checklist:

### DIFF
--- a/sdk/filter/result/filter_result.go
+++ b/sdk/filter/result/filter_result.go
@@ -1,8 +1,8 @@
 package result
 
 type KeyValueString struct {
-	key   string
-	value string
+	Key   string
+	Value string
 }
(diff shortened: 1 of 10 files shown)


PrajwalKrishna (approved):

varkey98:
I've updated the 2 server instrumentations  we have on goagent as well to use the updated filter result, pls let me know if we want to separate it onto a different pr

tim-mwangi (approved):

ryanericson (approved):

PR #239, opened 2024-10-07
feat: log opentelemetry logs and errors using a provided zap logger

varkey98:
## Description
As of now, the opentelemetry logs and errors are printed to stdout. Adding logic to use a passed zap logger for logging these. Opentelemetry go uses [go-logr](https://github.com/go-logr/logr) API for internal logging and go-logr has a zap [wrapper](https://github.com/go-logr/zapr) implementation which is used for passing the provided zap logger to the opentelemetry library.

### Testing
Please describe the tests that you ran to verify your changes. Please summarize what did you test and what needs to be tested e.g. deployed and tested helm chart locally. 

### Checklist:


varkey98:
error screenshot:
image

varkey98:
looking into static analysis failures

PrajwalKrishna (approved):

instrumentation/opentelemetry/errorhandler/handler.go
```diff
@@ -0,0 +1,22 @@
+package errorhandler
```
tim-mwangi, line 1: Can you check if this works for our custom batchspanprocessor or if it can be adapted the new zap logger? https://github.com/hypertrace/goagent/tree/main/instrumentation/opentelemetry/batchspanprocessor . We add some custom code in our batchspanprocessor like catching panics and metrics for spans received, sent and dropped. Since we cannot use the logger in the internal package in the otel go repo, we adapt it and even initialize a new stdout logger.

instrumentation/opentelemetry/init.go
```diff
@@ -245,6 +248,19 @@ func Init(cfg *config.AgentConfig) func() {
 // and returns a shutdown function to flush data immediately on a termination signal.
 func InitWithSpanProcessorWrapper(cfg *config.AgentConfig, wrapper SpanProcessorWrapper,
 	versionInfoAttrs []attribute.KeyValue) func() {
+	logger, err := zap.NewProduction()
```
tim-mwangi, line 251: Is there existing log config we can use to set some config in the zap logger we create?

instrumentation/opentelemetry/errorhandler/handler.go
varkey98: I didnt know we had a logger here. We can do it, zap has a global api. We can either wrap it in go-logr or use zap itself in the custom bsp

varkey98: done

instrumentation/opentelemetry/init.go
varkey98: We dont have any existing log config in the hypertrace proto. The traceable goagent initializes logger with this config: https://github.com/Traceableai/goagent-src/blob/main/init.go#L29. Maybe we can move that logic here? Or can just use the new zap compatible api there?

tim-mwangi (approved):

varkey98:
One caveat is that if any of our agents set a zap global logger, that will be overwritten during goagent initialization. I dont think any of our agents do this, but we should remember this in the future

instrumentation/opentelemetry/init.go
tim-mwangi: There's a techdebt ticket to move the hypertrace goagent login into traceable goagent and archive hypertrace goagent. So we should probably wait until we do that and in hypertrace we can go with some zap logger support. But it should be overridable by the traceable goagent one for now. I did not know that we already had a zap logger in traceable goagent.

varkey98: Oh okay, then I think we should be good. We can pass that zap logger here with the new api

puneet-traceable: My ask here is that we should be able to use the logger created in the agent(i.e. mirroring-agent/ebpf) so that we can control where do the logs go(file/stdout/stderr), loglevels, rotation from one place.

varkey98: Yes. As of now ebpf calls this api: https://github.com/Traceableai/ebpf/blob/main/pkg/export/exporter.go#L41
This calls the InitWithSpanProcessorWrapper here https://github.com/Traceableai/goagent-src/blob/main/init.go#L37. Now we'll update the goagent-src to also expose an api for taking in a zap logger, then we can update the respective clients (ebpf/mirroring-agent/ext_cap)

instrumentation/opentelemetry/batchspanprocessor/logger.go
```diff
@@ -2,35 +2,48 @@ package batchspanprocessor // import "github.com/hypertrace/goagent/instrumentat
...
 // Not an error but is likely more important than an informational event.
 func Warn(msg string, keysAndValues ...interface{}) {
-	logger.V(1).Info(msg, keysAndValues...)
+	getLogger().V(1).Info(msg, keysAndValues...)
+}
+
+// since global variables init happens on package imports,
+// the zap global logger might not be set by then
+// hence doing a lazy initialization
+func getLogger() logr.Logger {
```
puneet-traceable, line 44: Should we instead pass logger here https://github.com/hypertrace/goagent/blob/94709dab7c096106f9a8a6777b9400299657f4b2/instrumentation/opentelemetry/init.go#L309 and store that in batch_span process struct? What I am unsure is what this global logger is? Secondly if intend to pass a sampling logger from the agent, how would that work? Does it apply rotation that's configured?

instrumentation/opentelemetry/batchspanprocessor/logger.go
```diff
@@ -2,35 +2,48 @@ package batchspanprocessor // import "github.com/hypertrace/goagent/instrumentat
...
 func Debug(msg string, keysAndValues ...interface{}) {
-	logger.V(8).Info(msg, keysAndValues...)
+	getLogger().V(8).Info(msg, keysAndValues...)
 }
 
 // Warn prints messages about warnings in the API or SDK.
 // Not an error but is likely more important than an informational event.
 func Warn(msg string, keysAndValues ...interface{}) {
-	logger.V(1).Info(msg, keysAndValues...)
+	getLogger().V(1).Info(msg, keysAndValues...)
```
puneet-traceable, line 38: Also, do we need separate verbosity levels here? How does that gets configured?

varkey98: Not sure about this, kept the existing behavior. cc: @tim-mwangi

instrumentation/opentelemetry/batchspanprocessor/logger.go
varkey98: This works with whatever logger we pass as of now. I've checked with file logger on TPA. Because we set the global logger here: https://github.com/hypertrace/goagent/blob/main/instrumentation/opentelemetry/init.go#L288. Only case this wont work is when someone sets another zap logger from another go routine. But I dont think we do that anywhere. I didnt want to set it in the bsp struct because that'll modify the struct compared to otel and thought we didnt want to do that. @tim-mwangi ?

varkey98: traceable goagent has it's own global logger implemented with zap, I thought that when we port this to that repo, we can change the bsp to our global logger instead of zap global logger and this as makeshift mechanism for the time being

instrumentation/opentelemetry/batchspanprocessor/logger.go
tim-mwangi: This is based on the adaptation from https://github.com/open-telemetry/opentelemetry-go/blob/main/internal/global/internal_logging.go. I believe they are the levels in the logr package. If we are somehow hooking zap under the logr package, I think they should be compatible?

instrumentation/opentelemetry/init.go
tim-mwangi: Yeah. If the same zap logger or zapcore object can be used throughout, that would be ideal.

PR #240, opened 2024-10-21
feat: Support environment-service pair to map to traceproviders

thugrock7:


instrumentation/opentelemetry/init_additional.go
```diff
@@ -13,6 +13,11 @@ import (
 	"go.opentelemetry.io/otel/sdk/trace"
 )
 
+const (
+	environmentKey     = "deployment.environment"
```
varkey98, line 17: Since this is the key input from gateway agents, I think it should be prefixed with traceableai. I'd say traceableai.environment.name is a better key

varkey98 (commented):
Looking through the pr, I don't think we would need to change traceable or hypertrace goagent APIs. Can't we just manage this from TPA itself? cc @tim-mwangi @ryanericson ?

instrumentation/opentelemetry/init_additional.go
thugrock7: Wanted to keep "traceable" name out of hypertrace repos. Don't know the convention we follow though

varkey98:
Another thing is that all of these api's written assuming that service name is a resource attribute. The moment we start piggy backing on them for treating environment similarly to service name there'll be a disparity as TPA processors assume them to be a span attribute. This will happen across agents and easiest thing would be to use attributeaction processor to put this under both span and resource attributes till we update processors

instrumentation/opentelemetry/init_additional.go
varkey98: I'd rather keep environment notion itself out of this repo. It'll be scattered  if you add that notion here. Because initially we made it traceable specific one. ref, https://github.com/Traceableai/agent-config/blob/main/proto/ai/traceable/agent/config/v1/config.proto#L22

varkey98: Now it's gonna be like hypertrace agent provides environment specific tracers, but traceable goagent is used for getting environment specific filters when you think about it from a tracer-filter pair pov

thugrock7: I agree to keep environment notion out of this repo. Just thinking on lines of making traceable-goagent to have multi environment support as the TA, to register services there we mitigate that logic to `RegisterServiceWithSpanProcessorWrapper()` ([here](https://github.com/hypertrace/goagent/blob/main/instrumentation/opentelemetry/init.go#L396)), the only problem is we assume a unique key for traceProviders here which service name used to do. But if we want to create a multi-env support for traceable-goagent, where a same service name can exist in different environments, this current setup fails due to [this](https://github.com/hypertrace/goagent/blob/main/instrumentation/opentelemetry/init.go#L409) traceProviderMap. So here I only intend to use unique key to create tracer.
May be other way is to form a unique key in traceable-goagent itself, and use that to call the RegisterServiceWithSpanProcessorWrapper(unique_key, resourceAttributes), but as [here](https://github.com/hypertrace/goagent/blob/main/instrumentation/opentelemetry/init.go#L427) we add resource attrs, didn't feel it generic enough and dropped that idea

varkey98: I still dont understand why you can't add all of this logic in TPA itself?

varkey98: It's just that the argument name is service_name across our api, but actually we would be passing `environment::sep::service_name`. I dont think that'd be a problem, and it would be easier to change just that name than spreading out the logic imo

tim-mwangi:
Hey @thugrock7, what is this needed for? Do we want to support multiple environments in tracers as well?

thugrock7:
I was trying to make multi-environment support for ext_cap of TPA, which followed to make traceable-goagent to support multiple environments. It's different to other TA like ebpf, where TA can only one have environment.

instrumentation/opentelemetry/init_additional.go
thugrock7: > I still dont understand why you can't add all of this logic in TPA itself?

Currently we have for each service in extcap a [unique](https://github.com/Traceableai/traceable-agent/blob/main/ext_cap/internal/ext_cap/tracer_set.go#L15) tracerfilterpair .The tracerSets of ext_cap([link](https://github.com/Traceableai/traceable-agent/blob/main/ext_cap/internal/ext_cap/tracer_set.go#L88)) use goagent to create a tracer, which in current setup will fail for cases of same service name which can be case in multiple environments. So to avoid depending on only service name as key which can no longer be unique, formed the unique key here (encode) itself in PR. Although this could could come from TPA itself, then we should decode for place like [here](https://github.com/hypertrace/goagent/blob/main/instrumentation/opentelemetry/init.go#L427) to get to add resource attrs, and get service name etc correctly.

varkey98: why do you want to decode the value? Afair, the map itself would be coming from TPA right?

PR #243, opened 2024-11-18
fix: one exporter to rule them all

tim-mwangi:
## Description
One exporter for all registered services. We are currently creating a new exporter every time a new service is registered which results in multiple grpc clients to the collector being created.

### Testing
Manual tests

### Checklist:
- [ ] My changes generate no new warnings

### DIFF
--- a/instrumentation/opentelemetry/init.go
+++ b/instrumentation/opentelemetry/init.go
@@ -5,7 +5,6 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
-	"go.uber.org/zap"
 	"log"
 	"net/http"
 	"os"
@@ -40,6 +39,7 @@ import (
 	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
 	"go.opentelemetry.io/otel/trace"
 	"go.opentelemetry.io/otel/trace/noop"
+	"go.uber.org/zap"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/resolver"
 )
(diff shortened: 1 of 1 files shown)


instrumentation/opentelemetry/init.go
```diff
@@ -141,15 +148,17 @@ func makeExporterFactory(cfg *config.AgentConfig) func() (sdktrace.SpanExporter,
...
 				cfg.GetReporting().GetEndpoint().GetValue(),
 				zipkin.WithClient(client),
 			)
+			return globalExporter, globalExporterErr
 		}
 	case config.TraceReporterType_LOGGING:
 		return func() (sdktrace.SpanExporter, error) {
 			// TODO: Define if endpoint could be a filepath to write into a file.
-			return stdouttrace.New(stdouttrace.WithPrettyPrint())
+			globalExporter, globalExporterErr = stdouttrace.New(stdouttrace.WithPrettyPrint())
```
ryanericson, line 160: Does this need to be thread safe? If yes, do we need to worry about races resulting in some dangling exporters or that’s not a big deal?

tim-mwangi: Maybe not since it would set up in the Init method. Unless all the consumers do not call Init. But I will shelf this one for now since I did not observe any remarkable improvement with one exporter vs multiple exporters.

ryanericson: Not opposed to shelving it, but why no improvement? This should help to reduce number of export connetions?

tim-mwangi: I didn't have time to verify that using lsof or some other networking tool. I suspect reducing clients does not mean reducing connections as well. The clients maybe fewer but the connection count may still be the same. I would want to verify it next.

tim-mwangi:
Closing this one

PR #244, opened 2025-02-05
ENG-56723: Add optional token as header, add specs

prodion23:
After working on the ext_cap -> goagent-src -> goagent changes I realized the previous approach wouldn't work because the exporterFactory is only inited once, after that the config is static & it's really just the `exporterFactory()` that is invoked and needs to be aware of the new header. 

I've updated the approach to instead add a `WithHeaders` option we can pass down to customize the exporter that will be created from factory.
This also removes the original question about hardcoding constant for the key, we'll be able to control headers from TPA, which may be nice from a maintenance perspective.

ORIGINAL
--
This optionally adds token header to otlpgrpc, otlphttp, or zipkin. 

**Q1:** Do we have an expected key defined? This is currently marked as TODO- in the code.

**Q2:** Do we want key hardcoded constant for the header key or do we want the key also passed through TPA? 

We already have a `Token` field in the config, if we want to customize key we will need to update config to allow a map of headers or add a TokenKey/TokenValue for ex.

Please lmk direction we want to go, this impl is straightforward; however, allowing header key specification would be more robust - but from my current understanding maybe unnecessary.

- Unit tests assert header/metadata is send via the appended opts.


instrumentation/opentelemetry/init.go
```diff
@@ -132,6 +132,12 @@ func makeMetricsExporterFactory(cfg *config.AgentConfig) func() (metric.Exporter
 }
 
 func makeExporterFactory(cfg *config.AgentConfig) func() (sdktrace.SpanExporter, error) {
+	additionalHeaders := make(map[string]string)
+	if cfg.Reporting.GetToken() != nil {
+		// TODO - confirm what the header key should be, is this defined yet for go -> local-coll
+		additionalHeaders["TODO-traceable-agent-token"] = cfg.GetReporting().GetToken().GetValue()
```
tim-mwangi, line 138: I think we should make this configurable? But we can have a reasonable default like `traceable-agent-token`

tim-mwangi: What do you think @ryanericson ?

tim-mwangi (approved):

tim-mwangi:
> This optionally adds token header to otlpgrpc, otlphttp, or zipkin.
> 
> **Q1:** Do we have an expected key defined? This is currently marked as TODO- in the code.
> 
> **Q2:** Do we want key hardcoded constant for the header key or do we want the key also passed through TPA?
> 
> We already have a `Token` field in the config, if we want to customize key we will need to update config to allow a map of headers or add a TokenKey/TokenValue for ex.
> 
> Please lmk direction we want to go, this impl is straightforward; however, allowing header key specification would be more robust - but from my current understanding maybe unnecessary.
> 
> * Unit tests assert header/metadata is send via the appended opts.

I think we can allow for configuration of the token header name but have a reasonable default.

instrumentation/opentelemetry/init.go
ryanericson: It's `traceableai-agent-token` cc @varkey98

varkey98: I dont think we need to make it configurable. I have 2 reasons:
1. We have already hardcoded the key (traceableai-agent-token) in TPA
2. None of the other keys(traceableai-environment-name, traceableai-service-name, etc) are configurable as of today

tim-mwangi: Ok. No need to be configurable.

tim-mwangi (approved):

varkey98 (approved):

instrumentation/opentelemetry/init_test.go
```diff
@@ -16,7 +11,17 @@ import (
 	"go.opentelemetry.io/otel/propagation"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/trace/noop"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/resolver"
+	"log"
```
tim-mwangi, line 17: nit: dependencies moved.

instrumentation/opentelemetry/init.go
```diff
@@ -51,14 +51,31 @@ var (
...
 	}
 )
 
+type ServiceOption func(*serviceOptions)
+
+type serviceOptions struct {
+	headers map[string]string
+}
+
+func WithHeaders(headers map[string]string) ServiceOption {
```
tim-mwangi, line 68: I like this.

tim-mwangi (approved):

tim-mwangi (approved):

instrumentation/opentelemetry/init.go
```diff
@@ -408,7 +454,7 @@ func RegisterServiceWithSpanProcessorWrapper(key string, resourceAttributes map[
 		return nil, noop.NewTracerProvider(), fmt.Errorf("key %v is already used for initialization", key)
 	}
 
-	exporter, err := exporterFactory()
+	exporter, err := exporterFactory(opts...)
```
varkey98, line 457: From what I understood, this lets us control the headers from the consumer code, nice. But I'm confused about one thing, the default exporter, it still won't have any token right? I think that's perfectly fine from our ext_cap use case, but still might be better to make the behavior consistent and initialize that also with this option if cfg.Reporting.Token is present.

varkey98 (approved):

instrumentation/opentelemetry/init.go
varkey98: +1

tim-mwangi (approved):

instrumentation/opentelemetry/init.go
```diff
@@ -51,14 +51,31 @@ var (
...
+	headers map[string]string
+}
+
+func WithHeaders(headers map[string]string) ServiceOption {
+	return func(opts *ServiceOptions) {
+		if opts.headers == nil {
+			opts.headers = make(map[string]string)
+		}
+		for k, v := range headers {
+			opts.headers[k] = v
```
puneet-traceable, line 74: nit: can probably use maps.Copy() [ref](https://pkg.go.dev/maps#Copy)

prodion23: TIL 👍

varkey98 (approved):

instrumentation/opentelemetry/init.go
puneet-traceable: Option pattern is a fairly popular pattern in go.
More such things can be read here: https://abhinavg.net/2022/12/06/designing-go-libraries/
This guy used to lead go team at uber.

PR #245, opened 2025-02-14
feat: Update span interface to add a new function which will return span id

stellarhuman:
## Description
Added a way to directly get the spanId, from the span that is created.

### DIFF
--- a/instrumentation/opentelemetry/span.go
+++ b/instrumentation/opentelemetry/span.go
@@ -112,6 +112,10 @@ func (s *Span) AddEvent(name string, ts time.Time, attributes map[string]interfa
 	s.Span.AddEvent(name, trace.WithTimestamp(ts), trace.WithAttributes(otAttributes...))
 }
 
+func (s *Span) GetSpanId() string {
+	return s.Span.SpanContext().SpanID().String()
+}
+
 func SpanFromContext(ctx context.Context) sdk.Span {
 	return &Span{trace.SpanFromContext(ctx)}
 }
--- a/instrumentation/opentelemetry/span_test.go
+++ b/instrumentation/opentelemetry/span_test.go
@@ -148,3 +148,11 @@ func TestLen(t *testing.T) {
 	// service.instance.id is added implicitly in StartSpan so 3 attributes will be present.
 	assert.Equal(t, 3, s.GetAttributes().Len())
 }
+
+func TestGetSpanID(t *testing.T) {
+	_, s, _ := StartSpan(context.Background(), "test_span", &sdk.SpanOptions{})
+	spanId := s.GetSpanId()
+	if len(spanId) == 0 {
+		t.Error("span id should not be empty")
+	}
+}
--- a/instrumentation/opentelemetry/span_test.go
+++ b/instrumentation/opentelemetry/span_test.go
@@ -152,7 +152,5 @@ func TestLen(t *testing.T) {
 func TestGetSpanID(t *testing.T) {
 	_, s, _ := StartSpan(context.Background(), "test_span", &sdk.SpanOptions{})
 	spanId := s.GetSpanId()
-	if len(spanId) == 0 {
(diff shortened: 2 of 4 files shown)


instrumentation/opentelemetry/span.go
```diff
@@ -112,6 +112,10 @@ func (s *Span) AddEvent(name string, ts time.Time, attributes map[string]interfa
 	s.Span.AddEvent(name, trace.WithTimestamp(ts), trace.WithAttributes(otAttributes...))
 }
 
+func (s *Span) GetSpanId() string {
+	return s.Span.SpanContext().SpanID().String()
```
thugrock7, line 116: do we expect this value to be not null everytime or do need to something like

```
if (s.Span.SpanContext().SpanID().HasSpanID()) {
   s.Span.SpanContext().SpanID().String()
 }
 return ""
```

instrumentation/opentelemetry/span_test.go
```diff
@@ -148,3 +148,11 @@ func TestLen(t *testing.T) {
 	// service.instance.id is added implicitly in StartSpan so 3 attributes will be present.
 	assert.Equal(t, 3, s.GetAttributes().Len())
 }
+
+func TestGetSpanID(t *testing.T) {
+	_, s, _ := StartSpan(context.Background(), "test_span", &sdk.SpanOptions{})
+	spanId := s.GetSpanId()
+	if len(spanId) == 0 {
```
thugrock7, line 155: use assertions instead. ex: `assert.Equal(t, 8, SpanIDNoOfBytes)`

varkey98:
Do we really need this change here in goagent? We can simply do this in TPA if this is the route you want to go
```
	spanCtx := trace.SpanContextFromContext(ctx)
	if spanCtx.HasTraceID() {
		fmt.Println(spanCtx.TraceID().String())
	}

	if spanCtx.HasSpanID() {
		fmt.Println(spanCtx.SpanID().String())
	}
```

Adding it here is not really required imo

thugrock7:
Yes, functionally we can get this done using `trace.SpanContextFromContext(ctx)` as well, but thinking about this logically , as span ID is property of span, I simply want to do something like GetSpanID() on span to fetch this ID rather than knowing internals to call spanContext ourselves and get this done where ever span ID is required .

thugrock7:
your inputs here @ryanericson ?

ryanericson:
> > Yes, functionally we can get this done using `trace.SpanContextFromContext(ctx)` as well, but thinking about this logically , as span ID is property of span, I simply want to do something like GetSpanID() on span to fetch this ID rather than knowing internals to call spanContext ourselves and get this done where ever span ID is required .
> 
> Fair enough, but I'm not sure which is better

Ah I see what @varkey98 found which is kind of equivalent. But I agree with @thugrock7, just exposing it in our SDK is cleaner and I don't see any harm.

instrumentation/opentelemetry/span.go
stellarhuman: idts, if a span doest exist, you should always be able to call this method

varkey98: this should be fine. String's null value is empty string in golang. As long as it returns a string, we dont have to check here explicitly

varkey98 (approved):
LGTM

sdk/span.go
```diff
@@ -36,6 +36,8 @@ type Span interface {
 
 	// AddEvent adds an event to the Span with the provided name, timestamp and attributes.
 	AddEvent(name string, ts time.Time, attributes map[string]interface{})
+
+	GetSpanId() string
```
thugrock7, line 40: can you please add comment, explaining the functionality (just to follow existing conventions)

thugrock7:
Merging this, as @stellarhuman needs to use these changes on TPA side PR and want this PR to be included in this month release.

PR #250, opened 2025-06-26
draft: adding filter eval for client spans

varkey98:
Not complete yet, have to add filter call for db clients as well

### DIFF
--- a/sdk/instrumentation/google.golang.org/grpc/client.go
+++ b/sdk/instrumentation/google.golang.org/grpc/client.go
@@ -2,15 +2,16 @@ package grpc // import "github.com/hypertrace/goagent/sdk/instrumentation/google
 
 import (
 	"context"
-	"github.com/hypertrace/goagent/sdk/filter"
-	"google.golang.org/grpc/status"
 	"strings"
 
 	"github.com/hypertrace/goagent/sdk"
+	codes "github.com/hypertrace/goagent/sdk"
+	"github.com/hypertrace/goagent/sdk/filter"
 	internalconfig "github.com/hypertrace/goagent/sdk/internal/config"
 	"github.com/hypertrace/goagent/sdk/internal/container"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
 )
 
 // WrapUnaryClientInterceptor returns an interceptor that records the request and response message's body
@@ -72,7 +73,11 @@ func WrapUnaryClientInterceptor(
 
 			fr := filter.Evaluate(span)
 			if fr.Block {
-				return status.Error(StatusCode(int(fr.ResponseStatusCode)), StatusText(int(fr.ResponseStatusCode)))
+				statusText := StatusText(int(fr.ResponseStatusCode))
+				statusCode := StatusCode(int(fr.ResponseStatusCode))
+				span.SetStatus(codes.StatusCodeError, statusText)
+				span.SetAttribute("rpc.grpc.status_code", statusCode)
+				return status.Error(statusCode, statusText)
 			} else if fr.Decorations != nil {
 				if md, ok := metadata.FromOutgoingContext(ctx); ok {
 					for _, header := range fr.Decorations.RequestHeaderInjections {
--- a/sdk/instrumentation/google.golang.org/grpc/client.go
+++ b/sdk/instrumentation/google.golang.org/grpc/client.go
@@ -2,6 +2,8 @@ package grpc // import "github.com/hypertrace/goagent/sdk/instrumentation/google
 
 import (
 	"context"
+	"github.com/hypertrace/goagent/sdk/filter"
+	"google.golang.org/grpc/status"
 	"strings"
 
 	"github.com/hypertrace/goagent/sdk"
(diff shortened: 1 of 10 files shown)


sdk/instrumentation/net/http/transport.go
```diff
@@ -58,6 +62,29 @@ func (rt *roundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 		req.Body = io.NopCloser(bytes.NewBuffer(body))
 	}
 
+	filterResult := rt.filter.Evaluate(span)
+	if filterResult.Block {
```
prodion23, line 66: is this for blocking client spans?

varkey98: Yes, since we are anyways calling libtraceable, thought we can add blocking support as well

varkey98: Agreed. But we also have DLP based blocking which is basically to prevent data leaks and was made for client spans only, but at that time none of our agents did blocking eval on client spans and hence Amex said they'll all of their egress under a proxy and instrument our agent there so that DLP based blocking will work. I was thinking that even if we dont turn it on today, we can add the blocking capability in code and later when we want to turn it on, we can enable blocking for client spans as well, since we have a config on libtraceable, which by default skips client spans for blocking eval.

varkey98: But if you think even having the capability makes our feature ambiguous, then definitely removing this

prodion23: Hmm, okay, but don't remove just because I don't think worth it -

i think worth identifying what we are trying to do..(I see you also did header injections) so is the purpose of these tickets to add sampling support, or add sampling support, header injections & blocking. 
(my main intention was to ensure feature parity in the lang agents.., if we are adding these here, we need to make sure we do them in other lang agents)

I'd think this is a product question, if we see value in blocking client spans then sure..

prodion23:
2 questions: 
what is use case for the db call blocking, is it the same case as http clients, DLP? 

& similarly can db spans be sampled today?

sdk/instrumentation/net/http/transport.go
tim-mwangi: Sampling makes sense for client spans. Blocking I am not so sure. Let's get clarity from PM.

varkey98: I have started a thread here: https://harness.slack.com/archives/C08SEG313E3/p1751005003984389

We kind of discussed it back when rolling dlp out, but since its a huge lift, never prioritised it. I am unsure if we should prioritise it now as well, but I thought I can add it here since its not a lot of extra changes from adding sampling support to get it supported in goagent.

varkey98:
> maybe 2 more questions: what is use case for the db call blocking, is it the same case as http clients, DLP?
> 
> & similarly can db spans be sampled today?

From a product perspective, I dont think dbclient spans blocking makes a lot of sense. Sampling still makes sense to me

tim-mwangi:
@varkey98 what's the ETA of this one?

varkey98:
@tim-mwangi I'll try to wrap it up tomorrow my daytime

tim-mwangi:
> @tim-mwangi I'll try to wrap it up tomorrow my daytime

Okay. Sounds good.

sdk/instrumentation/google.golang.org/grpc/client.go
```diff
@@ -60,6 +72,23 @@ func WrapUnaryClientInterceptor(delegateInterceptor grpc.UnaryClientInterceptor,
 				setAttributesFromRequestOutgoingMetadata(ctx, span)
 			}
 
+			fr := filter.Evaluate(span)
+			if fr.Block {
+				statusText := StatusText(int(fr.ResponseStatusCode))
```
tim-mwangi, line 77: statustext is going to be set to status code?

varkey98: we have a table mapping http status codes to correspinding status texts here: https://github.com/hypertrace/goagent/blob/main/sdk/instrumentation/google.golang.org/grpc/status.go#L7

varkey98: Copied this from our server instrumentation code: https://github.com/hypertrace/goagent/blob/main/sdk/instrumentation/google.golang.org/grpc/server.go#L112

examples/postgres-query/go.mod
```diff
@@ -6,7 +6,9 @@ toolchain go1.24.2
 
 replace github.com/hypertrace/goagent => ../..
 
-replace github.com/hypertrace/goagent/instrumentation/hypertrace/github.com/jackc/hyperpgx => ../../instrumentation/opentelemetry/github.com/jackc/hyperpgx
+replace github.com/hypertrace/goagent/instrumentation/hypertrace/github.com/jackc/hyperpgx => ../../instrumentation/hypertrace/github.com/jackc/hyperpgx
```
tim-mwangi, line 9: I need to check why we need this replace.

varkey98: one replace was always there pointing to the opentelemetry one's go module. When I added options pattern I added it only to the hypertrace wrapper (for other instrumentations like the grpc one we only had it for our hypertrace wrapper). So for using it I switched the replace from the otel one to hypertrace. But stranger part was that `replace` was not working transitively, hence needed to add both.

sdk/instrumentation/google.golang.org/grpc/client.go
tim-mwangi: Ok. It's fine for now.

examples/postgres-query/go.mod
tim-mwangi: Hopefully it does not give me trouble during the migration :)

varkey98: yes, 🤞

tim-mwangi (approved):

PR #252, opened 2025-07-14
ENGTAI-64543: expose withGrpcConn to allow exporters to share connection

prodion23:
## Description
Originally I tried implementing this by storing a single global exporter and then each new registered service tracer would have it's own in memory exporter. This performed essentially the same(marginal benefit ~1% in benchmarks..)

As an alternative, otel exposes a `WithGRPCConn` argument we can pass to each exporter. This keeps our implementation simpler and achieves essentially the same result. 

Since we need to have a reference to the shared connection the flow would be: 
```
// Create the config before calling 1 of the .Init functions
grpcConn, err := hypertrace.CreateGrpcConn(agentCfg)

// we need to handle closing the Conn, the underlying TraceProviders will not shut it down since they don't "own" the connection
defer grpcConn.Close()

flusher, err := hypertrace.InitWithSpanProcessorWrapperAndZap(...<all the args>, opentelemetry.WithGrpcConn(grpcConn)
...
```

In my benchmarks using the WithGrpcConn option with 10 services and sending 10k spans per service:
```
| Metric                   | Without Shared Connection | With Shared Connection |
|--------------------------|---------------------------|------------------------|
| **Performance**          |                           |                        |
| Total Time (ns/op)       | 21,856,828                | 19,969,818             |
| Span Creation (ns/op)    | 21,855,983                | 19,968,738             |
| Shutdown (ns/op)         | 6,009,575,057             | 6,007,935,632          |
| Connection Creation (ns) | N/A                       | 184.0                  |
| **Memory**               |                           |                        |
| Allocation (MB)          | 7.416                     | 11.51                  |
| Total Allocation (MB)    | 31.09                     | 32.61                  |
| System (MB)              | 38.83                     | 39.08                  |
| GC Count                 | 7.000                     | 6.000                  |
| **Network**              |                           |                        |
| Max Connection Count     | 0                         | 0                      |
| Network Received (MB)    | 10.49                     | 15.28                  |
| Network Sent (MB)        | 10.44                     | 15.20                  |
| **Scale**                |                           |                        |
| Total Services           | 10.00                     | 10.00                  |
| Total Spans              | 10,000                    | 10,000                 |
```
(I used the gopsutil pkg for network tests and was reporting to local collector, so I think that may be messing up network stats specificall)

### DIFF
--- a/instrumentation/opentelemetry/init.go
+++ b/instrumentation/opentelemetry/init.go
@@ -345,20 +345,20 @@ func createCaCertPoolFromFile(certFile string) *x509.CertPool {
 
 // Init initializes opentelemetry tracing and returns a shutdown function to flush data immediately
 // on a termination signal.
-func Init(cfg *config.AgentConfig) func() {
-	return InitWithSpanProcessorWrapper(cfg, nil, versionInfoAttributes)
+func Init(cfg *config.AgentConfig, opts ...ServiceOption) func() {
+	return InitWithSpanProcessorWrapper(cfg, nil, versionInfoAttributes, opts...)
 }
 
 // InitWithSpanProcessorWrapper initializes opentelemetry tracing with a wrapper over span processor
 // and returns a shutdown function to flush data immediately on a termination signal.
 func InitWithSpanProcessorWrapper(cfg *config.AgentConfig, wrapper SpanProcessorWrapper,
-	versionInfoAttrs []attribute.KeyValue) func() {
+	versionInfoAttrs []attribute.KeyValue, opts ...ServiceOption) func() {
 	logger, err := zap.NewProduction()
 	if err != nil {
 		logger = nil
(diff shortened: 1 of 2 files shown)


t-santoshsahu:
I think creating connection explicitly will be better. This is good. 
@prodion23 Does it have timeouts config? We do timeouts on TPA, so this may be not needed for laodbalancing.

tim-mwangi (approved):

prodion23:
@t-santoshsahu 
OTLP grpc export does have a timeout option, though we weren't exposing it before(default of 10s) 
**however**; I don't think it is respected anyway because we are using the BatchSpanProcessor.

(default of 30s)
https://github.com/hypertrace/goagent/blob/0e09d2cd51468df1493de9e29e41a18e716aa181/instrumentation/opentelemetry/batchspanprocessor/batch_span_processor.modified.go#L264-L302

